PR #1, opened 2020-11-23
update to work with the simpler segment API

mschoch:

PR #2, opened 2020-11-27
Update to simpler segment api again

mschoch:


mschoch:
this is the same as https://github.com/blevesearch/zapx/pull/1 only without me introducing unnecessary merge conflict

PR #8, opened 2020-11-30
v14 remove plugin method

mschoch:
rather than have a method Plugin() which returns a value
implementing the interface (typed as segment.Plugin)

the code (in scorch) registering this version can instead
simply use &zapv15.ZapPlugin{} which satisfies the interface

this more naturally follows the Go convention of consumers
defining the interface they need to work with, and it
further simplifies the segment api package

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/RoaringBitmap/roaring v0.4.23
 	github.com/blevesearch/bleve_index_api v0.0.2
 	github.com/blevesearch/mmap-go v1.0.2
-	github.com/blevesearch/scorch_segment_api v0.0.3
+	github.com/blevesearch/scorch_segment_api v0.0.4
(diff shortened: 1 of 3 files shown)


sreekanth-cb:
Minor thing, the `&zapv15.ZapPlugin{}` in the commit-msg needs to be &zapv14.ZapPlugin{}?

sreekanth-cb (commented):
.

mschoch:
@sreekanth-cb so the same commit is cherry-picked across all zapx versions.  In hindsight I should not have used the specific version number in the commit message for this reason.  However, I think it's still reasonably clear what the commit-message is suggesting, so my vote is to not start over again just to change this.  What do you think?

mschoch:
OK, I think what I'll do is release the new version of scorch_segment_api, then there will additional commit to each of these branches with the go.mod changes.  Then I can do squash and merge, and rewrite the commit message then.

PR #9, opened 2020-11-30
v13 remove plugin method

mschoch:
rather than have a method Plugin() which returns a value
implementing the interface (typed as segment.Plugin)

the code (in scorch) registering this version can instead
simply use &zapv15.ZapPlugin{} which satisfies the interface

this more naturally follows the Go convention of consumers
defining the interface they need to work with, and it
further simplifies the segment api package


sreekanth-cb (commented):
Minor thing, the `&zapv15.ZapPlugin{}` in the commit-msg needs to be &zapv13.ZapPlugin{}?

PR #18, opened 2020-12-07
 Restore the mergeTermFreqNormLocsByCopying optimization (#44)

mschoch:
Even with custom or different chunk size/factors, it still
 looks safe and efficient to copy the freq,norm,locs bytes
 for saving the extra encodings of many uvarint's.

 Chunk factor differences across postings per segment
 doesn't look to affect the functionality here.

 By restoring this optimization, the performance of the merge
 paths improved by

    wildcard - 30%
    fuzzy2 - 34%

 for showfast tests.
 ref - https://issues.couchbase.com/browse/MB-41399

### DIFF
--- a/intDecoder.go
+++ b/intDecoder.go
@@ -103,6 +103,10 @@ func (d *chunkedIntDecoder) readUvarint() (uint64, error) {
 	return d.r.ReadUvarint()
 }
 
+func (d *chunkedIntDecoder) readBytes(start, end int) []byte {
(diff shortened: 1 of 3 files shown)


sreekanth-cb (approved):

mschoch:
I pulled this over from zap master, using `git format-patch`, and applied it with `git am`.  This is corresponds to: https://github.com/blevesearch/zap/pull/44

PR #19, opened 2020-12-08
Enabling configurable freq/norm

sreekanth-cb:
Enabling configurable freq/norm with storage
    
    With freq/norm disabled at the index layer, storage
    layer no longer needs to persist both freq/norm related
    information onto disk.
    A special frequency value of zero/0 is interpreted as
    a configuration for skipping the processing of norm
    value.
    
    So with freq/norm disabled, zap always store the freq
    value of 0 along with an LSB indicating the presence
    of location data.
    
    Savings are mostly of 1 byte per occurrence of the
    token as we always persist a freq value of 0 with
    the location LSB.

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -491,6 +491,7 @@ func mergeTermFreqNormLocsByCopying(term []byte, postItr *PostingsIterator,
 		}
 
 		newRoaring.Add(uint32(hitNewDocNum))
+
 		err = tfEncoder.AddBytes(hitNewDocNum, nextFreqNormBytes)
 		if err != nil {
 			return 0, 0, 0, err
@@ -537,8 +538,13 @@ func mergeTermFreqNormLocs(fieldsMap map[string]uint16, term []byte, postItr *Po
 
 		locs := next.Locations()
 
-		err = tfEncoder.Add(hitNewDocNum,
-			encodeFreqHasLocs(nextFreq, len(locs) > 0), nextNorm)
+		if nextFreq > 0 {
+			err = tfEncoder.Add(hitNewDocNum,
+				encodeFreqHasLocs(nextFreq, len(locs) > 0), nextNorm)
+		} else {
+			err = tfEncoder.Add(hitNewDocNum,
+				encodeFreqHasLocs(nextFreq, len(locs) > 0))
+		}
 		if err != nil {
 			return 0, 0, 0, nil, err
 		}
--- a/posting.go
+++ b/posting.go
@@ -361,6 +361,9 @@ func (i *PostingsIterator) readFreqNormHasLocs() (uint64, uint64, bool, error) {
 	}
 
 	freq, hasLocs := decodeFreqHasLocs(freqHasLocs)
+	if freq == 0 {
+		return freq, 0, hasLocs, nil
+	}
 
 	normBits, err := i.freqNormReader.readUvarint()
 	if err != nil {
@@ -380,9 +383,14 @@ func (i *PostingsIterator) skipFreqNormReadHasLocs() (bool, error) {
 		return false, fmt.Errorf("error reading freqHasLocs: %v", err)
 	}
 
+	freq, hasLocs := decodeFreqHasLocs(freqHasLocs)
+	if freq == 0 {
+		return hasLocs, nil
+	}
+
 	i.freqNormReader.SkipUvarint() // Skip normBits.
 
-	return freqHasLocs&0x01 != 0, nil // See decodeFreqHasLocs() / hasLocs.
+	return hasLocs, nil // See decodeFreqHasLocs() / hasLocs.
 }
 
 func encodeFreqHasLocs(freq uint64, hasLocs bool) uint64 {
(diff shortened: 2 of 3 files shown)


merge.go
```diff
@@ -484,16 +484,18 @@ func mergeTermFreqNormLocsByCopying(term []byte, postItr *PostingsIterator,
 	lastDocNum uint64, lastFreq uint64, lastNorm uint64, err error) {
 	nextDocNum, nextFreq, nextNorm, nextFreqNormBytes, nextLocBytes, err :=
 		postItr.nextBytes()
-	for err == nil && len(nextFreqNormBytes) > 0 {
+	for err == nil {
```
abhinavdangeti, line 487: Not sure I follow this entirely - why change this?

sreekanth-cb: Now that we can have optional freqNorm bytes independent of locationBytes, we can't skip the merging/copying of the location bytes even if there are no freqNorm bytes configured/present on the disk.

Until now, zap always had some non-nil `nextFreqNormBytes` whenever there are any valid location bytes present.  ie location data was optional but freqNorm data wasn't.
Due to this - many of the code flow has that assumption while making disk-level iterator advancements. This is one such case.

sreekanth-cb: Now that we can have optional freqNorm bytes independent of locationBytes, we can't skip the merging/copying of the location bytes even if there are no freqNorm bytes configured/present on the disk.

Until now, zap always had some non-nil nextFreqNormBytes whenever there are any valid location bytes present. ie location data was optional but freqNorm data wasn't.
Due to this - many of the code flow has that assumption while making disk-level iterator advancements. This is one such case.

abhinavdangeti (approved):

posting.go
```diff
@@ -355,6 +355,12 @@ func (i *PostingsIterator) readFreqNormHasLocs() (uint64, uint64, bool, error) {
 		return 1, i.normBits1Hit, false, nil
 	}
 
+	// check if freq/norm is disabled.
+	if i.freqNormReader.isNil() {
+		// locations could be there even without freq/norm.
```
mschoch, line 360: it's not clear to me from this comment, that this case is handled, or if this comment is worrying about that possibility

mschoch: also, it seems I'm late to the game, but the `isNil()` method doens't feel very idiomatic

mschoch (commented):
Hmmn, still need to think about part of this.

Also, what is the effective space savings?  It seems like we still need to record frequency of 0.  Is that a varint?  So, do we only save not writing out the norm?

posting.go
sreekanth-cb: When there is no freq/norm info available, then the readFreqNormHasLocs () call just does a NOOP and let the callers handle any further processing of location details is the intent here.

isNil()  was just using an already existing method on the intDecoder.

sreekanth-cb: > Hmmn, still need to think about part of this.
> 
> Also, what is the effective space savings? It seems like we still need to record frequency of 0. Is that a varint? So, do we only save not writing out the norm?

When disabled, we don't write any freq/norm info to disk except the `termNotEncoded` offset value here- https://github.com/blevesearch/zapx/blob/master/intcoder.go#L160

Yes, those are varints - https://github.com/blevesearch/zapx/blob/master/merge.go#L618

sreekanth-cb: Savings were roughly like 10% or so in a default index of travel-sample with hardcoded freq/norm disabled with all other default options kept as it is.

posting.go
```diff
@@ -493,29 +501,39 @@ func (i *PostingsIterator) nextAtOrAfter(atOrAfter uint64) (segment.Posting, err
...
+			rv.locs = i.nextSegmentLocs[:0]
 		}
-		rv.locs = i.nextSegmentLocs[:0]
 
 		numLocsBytes, err := i.locReader.readUvarint()
 		if err != nil {
 			return nil, fmt.Errorf("error reading location numLocsBytes: %v", err)
 		}
 
 		j := 0
```
mschoch, line 521: @sreekanth-cb this last part isn't clear to me.  it seems to only be relating to locations, and not conditional on freq/norm.  can you explain why this is needed?

mschoch (commented):
You asked me to take special care in this review.  It all seemed straightforward to me except that last section.  Were there any other specific concerns you had?

posting.go
sreekanth-cb: Now that rv.freq can have a zero value, the initializations of those `i.nextLocs` slice won't happen with skip freq/norm configs.
The below code is a workaround to sail through the case where the freq is 0 and we need single allocations of Location{} object for as many location details as we tread further.

line no: 528 nextLoc = &Location{}

sreekanth-cb:
> You asked me to take special care in this review. It all seemed straightforward to me except that last section. Were there any other specific concerns you had?

nothing specific, just that I was doing some context switches so had that feeling of incompleteness somewhere.

posting.go
mschoch: I see, so in the past, we used freq as a hint for the size to preallocate the number of location structs.  It seems that was always a bit unsafe wasn't it?  It seems the new code is generally safer since we're reading an arbitrary number of locations, the old code was trusting it to be right.

sreekanth-cb: yes. that is my thinking here too.

mschoch (approved):
Approving because it seems that in all cases, if freq>0 we do what we always did before.  So only users setting skip freq/norm option into the new behavior.  And since it isn't a file format change, if we find bug to fix later, it's not as catastrophic.

PR #38, opened 2020-12-18
Add ci lint

mschoch:


mschoch:
NOTE: this is a much more paired down set of linters than the API packages got, because our code is much messier here.  However, the point goal would be to enforce what we can now, and improve over time.

abhinavdangeti (approved):

PR #39, opened 2020-12-18
add lint config

mschoch:


mschoch:
This was partially to test whether or not CI was working for PRs on all branches.  It would appear that it is not :(

PR #73, opened 2021-06-28
Implement io.WriterTo interface

sreekanth-cb:
-which helps to persist a memory segment into
 the given writer.

### DIFF
--- a/build.go
+++ b/build.go
@@ -16,6 +16,8 @@ package zap
 
 import (
 	"bufio"
+	"fmt"
+	"io"
 	"math"
 	"os"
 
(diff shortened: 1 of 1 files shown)


mschoch:
Why can't we use the standard library [io.WriterTo](https://golang.org/src/io/io.go?s=6763:6830#L185)?  We would have to refactor the closing to happen outside zap, but that makes zaps usage even cleaner anyway.

build.go
```diff
@@ -46,40 +58,58 @@ func PersistSegmentBase(sb *SegmentBase, path string) error {
...
 	if err != nil {
 		cleanup()
 		return err
 	}
 
-	err = f.Sync()
+	return err
+}
+
+type bufWriter struct {
```
mschoch, line 82: I thought maybe we could use the existing CountHashWriter, but I guess the cost of computing unnecessary crc32 is too high....

mschoch (approved):
I think I would have organized the buffering/flushing stuff differently, but I don't have a concrete proposal to improve it...

abhinavdangeti (approved):

PR #95, opened 2021-11-30
PostingsIterator: Address nil pointer dereferencing

abhinavdangeti:
+ If the postingsList of the iterator is nil or empty,
  there's nothing to find, so we need not invoke the
  nextDocNumAtOrAfterClean(..) method of PostingsIterator
  when the iterator's postings is nil.

+ Related to: blevesearch/bleve#1606

### DIFF
--- a/posting.go
+++ b/posting.go
@@ -562,7 +562,12 @@ func (i *PostingsIterator) nextDocNumAtOrAfter(atOrAfter uint64) (uint64, bool, 
 		return 0, false, nil
 	}
 
-	if i.postings == nil || i.postings.postings == i.ActualBM {
+	if i.postings == nil || i.postings == emptyPostingsList {
+		// couldn't find anything
+		return 0, false, nil
+	}
+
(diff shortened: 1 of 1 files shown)


sreekanth-cb:
Let's discuss this in the next sync-up.
Though i.postings == nil check makes sense, and if that was the issue there, not clear about the stack trace crash dumps.

abhinavdangeti:
Updated the commit message (with the latest patch set) to indicate the exact stack trace that this commit attempts to address.

Note that https://github.com/blevesearch/bleve/issues/1606 captures several stack traces.

sreekanth-cb:
If postings was nil or chunkSize was zero, then it should have crashed at line no: 677 itself?

https://github.com/blevesearch/zapx/blob/master/posting.go#L677

So, not sure how this fix is going to address that? Or am I missing the obvious here.

sreekanth-cb (commented):
.

sreekanth-cb (approved):

PR #96, opened 2021-12-14
Move all iterator forward until a match, only if possible

abhinavdangeti:
Intend to address this panic ..

```
runtime.errorString: runtime error: invalid memory address or nil pointer dereference
  File "github.com/blevesearch/zapx/v15@v15.3.1/posting.go", line 595, in (*PostingsIterator).nextDocNumAtOrAfter
  File "github.com/blevesearch/zapx/v15@v15.3.1/posting.go", line 476, in (*PostingsIterator).nextAtOrAfter
  File "github.com/blevesearch/zapx/v15@v15.3.1/posting.go", line 471, in (*PostingsIterator).Advance
  File "github.com/blevesearch/bleve/v2@v2.2.1/index/scorch/snapshot_index_tfr.go", line 152, in (*IndexSnapshotTermFieldReader).Advance
  File "github.com/blevesearch/bleve/v2@v2.2.1/search/searcher/search_term.go", line 105, in (*TermSearcher).Advance
  ...
```

.. highlighted here: https://github.com/blevesearch/bleve/issues/1606#issuecomment-993150219

### DIFF
--- a/posting.go
+++ b/posting.go
@@ -568,7 +568,7 @@ func (i *PostingsIterator) nextDocNumAtOrAfter(atOrAfter uint64) (uint64, bool, 
 
 	i.Actual.AdvanceIfNeeded(uint32(atOrAfter))
 
-	if !i.Actual.HasNext() {
+	if !i.Actual.HasNext() || !i.all.HasNext() {
 		// couldn't find anything
 		return 0, false, nil
 	}
@@ -590,6 +590,10 @@ func (i *PostingsIterator) nextDocNumAtOrAfter(atOrAfter uint64) (uint64, bool, 
 			if err != nil {
 				return 0, false, err
 			}
+		}
+
+		if !i.all.HasNext() {
+			return 0, false, nil
 		}
 
 		allN = i.all.Next()


posting.go
```diff
@@ -583,7 +583,7 @@ func (i *PostingsIterator) nextDocNumAtOrAfter(atOrAfter uint64) (uint64, bool,
 	// n is the next actual hit (excluding some postings), and
 	// allN is the next hit in the full postings, and
 	// if they don't match, move 'all' forwards until they do
-	for allN != n {
+	for allN != n && i.all.HasNext() {
```
sreekanth-cb, line 586: If you check the code for Next() then it internally has a call for hasNext().  (Thinking if Next() crashes then HasNext() can crash as well).
So, I don't think this explicit check would fix the crash.

Rather, the former PR which exits early upon a "i.postings == nil || i.postings == emptyPostingsList" would fix/improve the situation seems more intuitive.

abhinavdangeti: > If you check the code for Next() then it internally has a call for hasNext(). (Thinking if Next() crashes then HasNext() can crash as well).

That's not accurate though. Next() dereferences a pointer before it calls hasNext() (whose signature differs from HasNext()). We always check i.Actual's HasNext() before calling Next() but it isn't the same for i.all.

sreekanth-cb (commented):
Any explanation for - why it didn't crash on the first dereferencing or call to all.Next() at line no: 577 (https://github.com/blevesearch/zapx/blob/master/posting.go#L577)

A few of the hasNext/HasNext implementations I saw were just checking the pos variable against the container size. 
Can please you point me to the HasNext() implementation checking the sanity of the iterator/iter variable? (As that would prevent this crash is our expectation here)

I am not yet comfortable with the change as I am not following the situation in mind. 

Another thing is - if we aren't able to advance all pointers together, then we ought to error out than just coming out of the for loop and continuing the remaining flow?

abhinavdangeti:
So I'm not obviously familiar with the internals of the roaring, but it looks to me like that the stack trace from the issue occurs after a few iterations of the for loop where all.Next() is invoked. 

Perhaps we should be calling HasNext() every time before Next() is called. Just trying to keep the `all` iterator consistent with `iterator`.

sreekanth-cb (commented):
Looks better.

If we aren't able to advance all pointers together, then we ought to error out than just coming out of the for loop(under the assumption that all pointers are advanced enough) and let the remaining flow proceed?

sreekanth-cb (approved):

PR #107, opened 2022-02-22
Place bounds check within memUvarintReader's ReadUvarint

abhinavdangeti:
This bounds check is to return an empty uint64 while
decoding a 64-bit integer, should the reader have already
read all the bytes available.

For: https://github.com/blevesearch/bleve/issues/1651

```
Some context on the memUvarintReader ..

The code here reflects ReadUVarint(..) from
https://cs.opensource.google/go/go/+/refs/tags/go1.13.7:src/encoding/binary/varint.go

But as reported here: https://github.com/golang/go/issues/40618,
it appears it contained an error prone path that was addressed in 1.15.
```

### DIFF
--- a/memuvarint.go
+++ b/memuvarint.go
@@ -39,6 +39,11 @@ func (r *memUvarintReader) Len() int {
 // ReadUvarint reads an encoded uint64.  The original code this was
 // based on is at encoding/binary/ReadUvarint().
 func (r *memUvarintReader) ReadUvarint() (uint64, error) {
+	if r.C >= len(r.S) {
+		// nothing else to read
+		return 0, nil
+	}
+
 	var x uint64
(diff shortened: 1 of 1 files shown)


memuvarint.go
```diff
@@ -44,6 +44,11 @@ func (r *memUvarintReader) ReadUvarint() (uint64, error) {
 	var C = r.C
 	var S = r.S
 
+	if C >= len(S) {
```
sreekanth-cb, line 47: Minor thing - this check could be the first thing in this function so that lines 42-45 become redundant in such cases.

abhinavdangeti: Ack.

sreekanth-cb (approved):

PR #125, opened 2022-08-04
Disk stats computation bug fix

Thejas-bhat:
- Refactored the code with respect to disk stats computation, mainly the adding of chunks in content and int coders for location and doc values based information.
- Addressed a bug which was basically with respect to interim object maintaining state across New() calls.

### DIFF
--- a/new.go
+++ b/new.go
@@ -78,10 +78,13 @@ func (*ZapPlugin) newWithChunkMode(results []index.Document,
 		s.FieldsMap, s.FieldsInv, uint64(len(results)),
 		storedIndexOffset, fieldsIndexOffset, fdvIndexOffset, dictOffsets)
 
+	// get the bytes written before the interim's reset() call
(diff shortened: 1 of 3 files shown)


abhinavdangeti (approved):

new.go
```diff
@@ -51,6 +51,10 @@ func (*ZapPlugin) newWithChunkMode(results []index.Document,
 	chunkMode uint32) (segment.Segment, uint64, error) {
 	s := interimPool.Get().(*interim)
 
+	// this can be a reused interim object, so need to reset the
+	// bytes written stat count
+	s.setBytesWritten(0)
```
sreekanth-cb, line 56: This better placed inside the (s *interim) reset() method?

Thejas-bhat: Done

sreekanth-cb (approved):

PR #131, opened 2022-10-12
Making IO Stats Optional

Thejas-bhat:
It looks like the IO stats computation is causing too much of a performance overhead in terms of query throughput and latency. Making the stats computation optional until further optimisations are pushed in.

### DIFF
--- a/posting.go
+++ b/posting.go
@@ -255,15 +255,22 @@ func (p *PostingsList) Count() uint64 {
 // the bytes read from the postings lists stored
 // on disk, while querying
 func (p *PostingsList) ResetBytesRead(val uint64) {
-	atomic.StoreUint64(&p.bytesRead, val)
+	if AccountIOStats {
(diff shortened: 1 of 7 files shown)


posting.go
```diff
@@ -368,15 +372,19 @@ func (i *PostingsIterator) Size() int {
 // the freqNorm and location specific information
 // of a hit
 func (i *PostingsIterator) ResetBytesRead(val uint64) {
-	atomic.StoreUint64(&i.bytesRead, val)
+	if AccountIOStats {
+		atomic.StoreUint64(&i.bytesRead, val)
+	}
 }
 
 func (i *PostingsIterator) BytesRead() uint64 {
```
abhinavdangeti, line 380: Worth adding the flag here and everywhere else for BytesRead and BytesWritten as well to avoid the atomic op.

abhinavdangeti (approved):

PR #119, opened 2022-07-15
Fix IO Stats computation

Thejas-bhat:
The newly introduced stats in the PR [#1702](https://github.com/blevesearch/bleve/pull/1702) and [#117](https://github.com/blevesearch/zapx/pull/117)
are currently computed all time the irrespective of whether or not its needed. Since it is in a fairly hot code path, it could affect the performance when these stats are not needed. Hence, this PR aims to hide those stats computation behind a flag which controls the same.

PR #147, opened 2022-11-19
Fix alignment in chunkedContentCoder

zeripath:
An unaligned atomic bug was unfortunately introduced in #119 because the `bytesWritten` field was placed at the end of the `chunkedContentCoder` struct.

This places this after the bytes.Buffers and the bool causing it to be misaligned.

The ideal placement of this variable is not entirely clear but placing it before the progresiveWrite bool should help.

An alternative would be to just place this atomic field at the top of the struct then there would be no risk of it becoming misaligned in future.

I moved a few things around to reduce the size of the struct too but it could be possible to adjust things a little more to make the struct a little smaller.

Signed-off-by: Andrew Thornton <art27@cantab.net>

### DIFF
--- a/contentcoder.go
+++ b/contentcoder.go
@@ -31,8 +31,10 @@ func init() {
 	reflectStaticSizeMetaData = int(reflect.TypeOf(md).Size())
 }
 
-var termSeparator byte = 0xff
-var termSeparatorSplitSlice = []byte{termSeparator}
+var (
+	termSeparator           byte = 0xff
+	termSeparatorSplitSlice      = []byte{termSeparator}
+)
 
 type chunkedContentCoder struct {
 	final     []byte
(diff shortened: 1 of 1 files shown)


Thejas-bhat:
Thanks you very much @zeripath for raising this bug. I'm not entirely sure about the specifics of the bug over here. However, it looks like the panic message that the user has posted in the issue https://github.com/go-gitea/gitea/issues/21869 shows that the atomic.AddUint64() is the cause of panic.
```
goroutine 1098 [running]:
runtime/internal/atomic.panicUnaligned()
        /usr/local/go/src/runtime/internal/atomic/unaligned.go:8 +0x2d
runtime/internal/atomic.Xadd64(0x1d210df4, 0x8)
        /usr/local/go/src/runtime/internal/atomic/atomic_386.s:125 +0x11
github.com/blevesearch/zapx/v15.(*chunkedContentCoder).incrementBytesWritten(...)
        /go/pkg/mod/github.com/blevesearch/zapx/v15@v15.3.5/contentcoder.go:114
```

Now, I also observe that the version of the zapx and bleve used is 15.3.5 and 2.3.4 respectively. These releases had the atomic operations in incrementBytesWritten method (which is why I'm assuming its failing). The upcoming release of bleve (2.3.6) will have the zapx tag 15.3.7 which avoids the atomic operations altogether. My thinking is that this issue won't occur when you use 2.3.6. However, if the issue still persists in that release, please go ahead create an issue or a PR (if you've verified a fix for the same) and add me as a reviewer for it. 

Thank you!
Thejas

abhinavdangeti (approved):

Thejas-bhat (approved):

PR #148, opened 2023-02-17
Fixed 64-bit alignment panic in chunkedContentCoder on 32-bit platforms.

cj123:
Following the fix in #147, a 'panic: unaligned 64-bit atomic operation' is still occurring on some platforms.

As noted in #147, this commit implements the alternative fix by placing the bytesWritten field at the top of the struct. Per the Golang [sync/atomic documentation](https://pkg.go.dev/sync/atomic#pkg-note-BUG), on ARM, 386 and 32-bit MIPS systems, this is the reliable way to prevent these issues.

### DIFF
--- a/contentcoder.go
+++ b/contentcoder.go
@@ -37,6 +37,8 @@ var (
 )
 
 type chunkedContentCoder struct {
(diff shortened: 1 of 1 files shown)


abhinavdangeti (approved):
Thanks for the contribution @cj123 . Looks good to me. Let's wait on one other review.

Thejas-bhat (approved):

PR #149, opened 2023-02-23
Place a ceiling for initial capacity allocated for bytes.Buffer

abhinavdangeti:
+ Initializing the ceiling to 10MB. This means that at least
   10 MB of space is guaranteed if the estimate is larger than
    that. Any additional space will be allocated on demand.
+ Intended to address: https://github.com/blevesearch/bleve/issues/1780

### DIFF
--- a/new.go
+++ b/new.go
@@ -32,6 +32,7 @@ import (
 var NewSegmentBufferNumResultsBump int = 100
 var NewSegmentBufferNumResultsFactor float64 = 1.0
 var NewSegmentBufferAvgBytesPerDocFactor float64 = 1.0
+var MaxSegmentBufferInitialSize int = 10 * 1024 * 1024 // 10MB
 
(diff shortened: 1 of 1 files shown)


new.go
```diff
@@ -32,6 +32,7 @@ import (
 var NewSegmentBufferNumResultsBump int = 100
 var NewSegmentBufferNumResultsFactor float64 = 1.0
 var NewSegmentBufferAvgBytesPerDocFactor float64 = 1.0
+var MaxSegmentBufferInitialSize int = 10 * 1024 * 1024 // 10MB
```
Thejas-bhat, line 35: I'm thinking, does it make it sense it make this as a config option in bleve? If we have that, perhaps it would make things a bit more flexible and accommodative based on the use case of application, like the application can run some iterations with different values and set it as a config which works the best for their use case which need not be the optimal value for other use cases.

CascadingRadium:
Not planned

PR #151, opened 2023-04-11
Update PostingsList's chunkSize only when err == nil

abhinavdangeti:
Possibly related to: https://github.com/blevesearch/bleve/issues/1811

### DIFF
--- a/posting.go
+++ b/posting.go
@@ -303,11 +303,16 @@ func (rv *PostingsList) read(postingsOffset uint64, d *Dictionary) error {
 		return fmt.Errorf("error loading roaring bitmap: %v", err)
 	}
 
-	rv.chunkSize, err = getChunkSize(d.sb.chunkMode,
+	chunkSize, err := getChunkSize(d.sb.chunkMode,
 		rv.postings.GetCardinality(), d.sb.numDocs)
 	if err != nil {
 		return err
+	} else if chunkSize == 0 {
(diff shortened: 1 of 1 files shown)


posting.go
```diff
@@ -303,12 +303,14 @@ func (rv *PostingsList) read(postingsOffset uint64, d *Dictionary) error {
 		return fmt.Errorf("error loading roaring bitmap: %v", err)
 	}
 
-	rv.chunkSize, err = getChunkSize(d.sb.chunkMode,
+	chunkSize, err := getChunkSize(d.sb.chunkMode,
```
Thejas-bhat, line 306: A doubt here, from the code:
```
	case chunkMode == 1026:
		// improve upon the ideas tested in chunkMode 1025
		// the observation that the fewest number of dense chunks is the most
		// desirable layout, given the built-in assumptions of chunking
		// (that we want to put an upper-bound on the number of items you must
		//  walk over without skipping, currently tuned to 1024)
		//
		// 1.  compute the number of chunks needed (max 1024/chunk)
		// 2.  convert to chunkSize, dividing into maxDocs
		numChunks := (cardinality / 1024) + 1
		chunkSize := maxDocs / numChunks

		return chunkSize, nil
```
it seems like we could potentially return a zero chunkSize if the passed value of d.sb.numDocs is zero. Do we handle that part as well?

abhinavdangeti: Ya good point.

Thejas-bhat (approved):

PR #159, opened 2023-07-05
Index sections

Thejas-bhat:
work in progress


section_inverted_index.go
```diff
@@ -0,0 +1,942 @@
...
+	return dictOffsets, nil
+}
+
+func (io *invertedIndexOpaque) process(field index.Field, fieldID uint16, docNum uint64) {
+	if !io.init && io.results != nil {
+		io.prepareDicts()
+		io.init = true
+	}
+
+	if fieldID == ^uint16(0) {
```
abhinavdangeti, line 589: Would you add some commentary around when fieldID is set to this - which should be math.MaxUint16.

segment.go
```diff
@@ -89,19 +98,21 @@ func (*ZapPlugin) Open(path string) (segment.Segment, error) {
...
+	mem                  []byte
+	memCRC               uint32
+	chunkMode            uint32
+	fieldsMap            map[string]uint16   // fieldName -> fieldID+1
+	fieldsInv            []string            // fieldID -> fieldName
+	fieldsSectionsMap    []map[uint16]uint64 // fieldID -> section -> address
+	numDocs              uint64
+	storedIndexOffset    uint64
+	fieldsIndexOffset    uint64
+	newFieldsIndexOffset uint64
```
abhinavdangeti, line 110: I suppose the reason you kept `newFieldsIndexOffset` along side `fieldsIndexOffset` is for backward compatibility? The new attribute could use a better name implying sections are in play.

segment.go
```diff
@@ -294,6 +312,69 @@ func (s *SegmentBase) loadFields() error {
 	return nil
 }
 
+func (s *SegmentBase) loadFieldsNew() error {
```
abhinavdangeti, line 315: Same here - name to indicate sections.

Or perhaps it'd be cleaner to declare a new struct which comes with new attributes and methods -
```
type SegmentBase struct {
    // leave as is
}

type SegmentBaseV2 struct {
    SegmentBase

    sectionFieldsIndexOffset uint64
    ...
}

func (s *SegmentBaseV2) loadSectionFields() error {

}

type Segment struct {
    SegmentBaseV2

    ...
}
```

If this can cause way too much of refactoring, you can ditch it. Thoughts?

section_inverted_index.go
Thejas-bhat: yeah, i'll be adding commentary to certain portions of the changes in the coming commits.

segment.go
Thejas-bhat: it was more related to the fact that the newFieldsIndexOffset relates to information that has fieldsIndex info tagged alongside the section type informations.

segment.go
Thejas-bhat: oh cool, thanks! I'll think more on these lines and get back to you on this.

Thejas-bhat:
duplicate of #165

PR #165, opened 2023-07-18
Index sections

Thejas-bhat:
- Introduces the concept of section which talks about reformating the index file as a set of sections each section corresponding to a certain type of index. 
- Refactors the inverted index data i.e. the term dictionary, postings list etc. as part of a section.


write.go
```diff
@@ -85,9 +136,9 @@ func persistFields(fieldsInv []string, w *CountHashWriter, dictLocs []uint64) (u
 
 // FooterSize is the size of the footer record in bytes
 // crc + ver + chunk + field offset + stored offset + num docs + docValueOffset
```
abhinavdangeti, line 138: Update comment above reflecting new offset below.

build.go
```diff
@@ -160,24 +160,32 @@ func persistStoredFieldValues(fieldID int,
 func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32,
 	fieldsMap map[string]uint16, fieldsInv []string, numDocs uint64,
 	storedIndexOffset uint64, fieldsIndexOffset uint64, docValueOffset uint64,
-	dictLocs []uint64) (*SegmentBase, error) {
+	dictLocs []uint64, sectionsIndexOffset uint64) (*SegmentBase, error) {
```
abhinavdangeti, line 163: I thought fieldsIndex, docValuesIndex and storedValuesIndex will all going to get "sectioned"? In which case - placing a "sectionsIndexOffset" alongside storedIndexOffset, fieldsIndexOffset, docValueOffset seems somewhat misleading?

I gather sectionsIndexOffset will point to that section that holds either the FST or the vector index.

build.go
```diff
@@ -160,24 +160,32 @@ func persistStoredFieldValues(fieldID int,
...
+		fieldDvReaders:      make([]map[uint16]*docValueReader, len(segmentSections)),
+		docValueOffset:      docValueOffset,
+		dictLocs:            dictLocs,
+		fieldFSTs:           make(map[uint16]*vellum.FST),
 	}
 	sb.updateSize()
 
-	err := sb.loadDvReaders()
+	// FIXME load map based on new fields index offset data
+	// (maybe this works now)
```
abhinavdangeti, line 182: Maybe? :)

section_inverted_index.go
```diff
@@ -0,0 +1,962 @@
+package zap
```
abhinavdangeti, line 1: Should probably call this file section_inverted_text_index.go .

section_inverted_index.go
```diff
@@ -0,0 +1,962 @@
...
+	"math"
+	"sort"
+
+	"github.com/RoaringBitmap/roaring"
+	index "github.com/blevesearch/bleve_index_api"
+	seg "github.com/blevesearch/scorch_segment_api/v2"
+	"github.com/blevesearch/vellum"
+)
+
+type invertedIndexSection struct {
```
abhinavdangeti, line 15: Rename to invertedTextIndexSection.

write.go
```diff
@@ -50,16 +50,16 @@ func writeRoaringWithLen(r *roaring.Bitmap, w io.Writer,
 	return tw, nil
 }
 
-func persistFields(fieldsInv []string, w *CountHashWriter, dictLocs []uint64) (uint64, error) {
+func persistFieldsSectionsIndex(fieldsInv []string, w *CountHashWriter, dictLocs []uint64, opaque map[int]resetable) (uint64, error) {
```
abhinavdangeti, line 53: persistFieldsSection should suffice for the naming here.

build.go
Thejas-bhat: i've trimmed down the method signatures, please let me know if they look good.

build.go
```diff
@@ -159,25 +159,33 @@ func persistStoredFieldValues(fieldID int,
...
-		fieldFSTs:         make(map[uint16]*vellum.FST),
+		mem:                 mem,
+		memCRC:              memCRC,
+		chunkMode:           chunkMode,
+		fieldsMap:           fieldsMap,
+		fieldsInv:           fieldsInv,
+		numDocs:             numDocs,
+		storedIndexOffset:   storedIndexOffset,
+		fieldsIndexOffset:   sectionsIndexOffset,
+		sectionsIndexOffset: sectionsIndexOffset,
```
abhinavdangeti, line 173: How come the fieldsIndexOffset and sectionsIndexOffset are the same here?

merge.go
```diff
@@ -74,14 +73,14 @@ func mergeSegmentBases(segmentBases []*SegmentBase, drops []*roaring.Bitmap, pat
 	// wrap it for counting (tracking offsets)
 	cr := NewCountHashWriterWithStatsReporter(br, s)
 
-	newDocNums, numDocs, storedIndexOffset, fieldsIndexOffset, docValueOffset, _, _, _, err :=
+	newDocNums, numDocs, storedIndexOffset, fieldsIndexOffset, docValueOffset, _, _, _, sectionsIndexOffset, err :=
 		MergeToWriter(segmentBases, drops, chunkMode, cr, closeCh)
```
abhinavdangeti, line 77: Let's try and keep the signature of InitSegmentBase and MergeToWriter similar in terms of the index offsets.

abhinavdangeti (commented):
Couple concerns around API signatures and offset expectations.

merge.go
Thejas-bhat: Did you mean `interm.convert()`'s signature over here?

build.go
Thejas-bhat: essentially sections index content contains the same as fields index + sections information. so, for the fieldsIndexOffset which gives info about the fields in a segment we just consult the sectionsIndexOffset to get that info. Also, the field is needed for backward compatibility purposes as well.

CascadingRadium:
brilliant piece of code 💯 🥇

segment.go
```diff
@@ -294,6 +311,83 @@ func (s *SegmentBase) loadFields() error {
...
+	pos += uint64(sz)
+
+	for sectionIdx := uint64(0); sectionIdx < fieldNumSections; sectionIdx++ {
+		// read section id
+		fieldSectionType := binary.BigEndian.Uint16(s.mem[pos : pos+2])
+		pos += 2
+		fieldSectionAddr := binary.BigEndian.Uint64(s.mem[pos : pos+8])
+		pos += 8
+		fieldSectionMap[fieldSectionType] = fieldSectionAddr
+		if fieldSectionType == sectionInvertedIndex {
```
CascadingRadium, line 369: bro like here we have dictLocs as []int
where idx is fieldID and value is the offset to the invertedSection.
but problem is there when u have 2 sections. like dictLocs[2] will always have offset to inverted section. `Maybe u can make dictLocs as [][]int where dictLocs[2][1] will be address of sectionID 1 for fieldID 2`

Thejas-bhat: yeah, dictLocs maps a fieldID to the starting offset of the term dictionary. the slice is populated only for sectionInvertedIndex, so the slice won't have any other section's information.

CascadingRadium: so for other section u use another field in segment struct? ok i will use new then thanks bro

metonymic-smokey (commented):
The unit test TestMergeWithEmptySegment seems to fail for me with these new changes - must be an issue on my end since it passes in CI

section.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/RoaringBitmap/roaring"
+	index "github.com/blevesearch/bleve_index_api"
+)
+
+type section interface {
+	// process is essentially parsing of a specific field's content in a specific
+	// document. any tracking of processed data *specific to this section* should
+	// be done in opaque which will be passed to the Persist() API.
+	Process(opaque map[int]resetable, docNum uint64, f index.Field, fieldID uint16)
```
CascadingRadium, line 26: docNum datatype should be uint32, because we use a uint32 roaring bitmap to track deleted documents, so we should keep the docNum datatype consistent.

Thejas-bhat: the main reason why i kept it as uint64 was because it was like that in the original code, however i think this makes sense to me.

Thejas-bhat:
the main reason why I kept the docNum's type to be uint64 was to ensure that we don't break anything around the existing code - I was being conservative to make additional changes to the existing function signatures etc. do you all think its better to finalize whether to use uint32 or uint64 for docNum (after testing it better for eg) in a later PR? or do you want me to include this in the current PR?

abhinavdangeti:
> the main reason why I kept the docNum's type to be uint64 was to ensure that we don't break anything around the existing code - I was being conservative to make additional changes to the existing function signatures etc. do you all think its better to finalize whether to use uint32 or uint64 for docNum (after testing it better for eg) in a later PR? or do you want me to include this in the current PR?

@Thejas-bhat I pushed that last commit up to unbreak the build. The type of the faissIndexSection's Process didn't match with the section interface's Process. We can talk about changing the type later if really needed.

abhinavdangeti (approved):

PR #173, opened 2023-10-12
updating merge path with the ivf index creations

abhinavdangeti:


section_vector_index.go
```diff
@@ -211,6 +211,64 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(fieldID int, sbs []*Segme
...
+		metric := vecIndexes[0].MetricType()
+		finalVecIDs := maps.Keys(vecToDocID)
+
+		index, err := faiss.IndexFactory(dims, "IDMap2,IVF100,SQ8", metric)
+		if err != nil {
+			return err
+		}
+
+		index, _ = index.AsIVF()
+		err = index.MakeDirectMap(2)
```
abhinavdangeti, line 242: This area needs fixing.

section_vector_index.go
```diff
@@ -211,6 +211,64 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(fieldID int, sbs []*Segme
...
+		dims := vecIndexes[0].D()
+		metric := vecIndexes[0].MetricType()
+		finalVecIDs := maps.Keys(vecToDocID)
+
+		index, err := faiss.IndexFactory(dims, "IDMap2,IVF100,SQ8", metric)
+		if err != nil {
+			return err
+		}
+
+		index = index.AsIVF()
```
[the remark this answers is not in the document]
abhinavdangeti: This is still incomplete.

section_vector_index.go
```diff
@@ -211,6 +211,73 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(fieldID int, sbs []*Segme
 		vecIndexes = append(vecIndexes, index)
 	}
 
+	if len(vecToDocID) > 10000 {
```
abhinavdangeti, line 214: Should perhaps make this configurable?

If you chose 10000 only because faiss advertises the flat indexes perform well up until 10000 - I'm thinking if we should do some testing on our side to decide what a better value for this could be.

PR #175, opened 2023-11-03
command line tool support for vector search

Thejas-bhat:
- also includes fixing the docvalues parsing and fields tool parsing given the new index sections format.

### DIFF
--- a/cmd/zap/cmd/docvalue.go
+++ b/cmd/zap/cmd/docvalue.go
@@ -28,7 +28,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func docValueCmd(data []byte, args []string, field string, id int, fieldEndLoc, fieldStartLoc uint64) error {
+func dumpDocValueResults(data []byte, args []string, field string, id int, fieldEndLoc, fieldStartLoc uint64) error {
 	var chunkOffsetsPosition, offset, numChunks uint64
 
 	var fieldName string
@@ -81,7 +81,10 @@ func docValueCmd(data []byte, args []string, field string, id int, fieldEndLoc, 
 		fieldDvSize = float64(dvSize) / (1024 * 1024)
 		fieldChunkLens = append(fieldChunkLens, chunkLens...)
 		fieldChunkCount = numChunks
+	} else {
+		return nil
 	}
+
 	mbsize := float64(totalDvSize) / (1024 * 1024)
 	fmt.Printf("Total Doc Values Size on Disk: %.6f MB\n", mbsize)
 
@@ -206,7 +209,7 @@ func loadFieldData(data []byte, pos uint64, fieldID uint64, sectionMap map[uint1
 	return fieldsInv, nil
 }
 
-func legacyDocValueCmd(cmd *cobra.Command, args []string) error {
+func docValueCmd(cmd *cobra.Command, args []string) error {
 	if len(args) < 1 {
 		return fmt.Errorf("must specify index file path")
 	}
@@ -253,7 +256,7 @@ func legacyDocValueCmd(cmd *cobra.Command, args []string) error {
 				" persisted \n", id, field, fieldStartLoc, fieldStartLoc)
 			continue
 		}
-		err = docValueCmd(data, args, field, id, fieldEndLoc, fieldStartLoc)
+		err = dumpDocValueResults(data, args, field, id, fieldEndLoc, fieldStartLoc)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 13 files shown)


cmd/zap/cmd/docvalue.go
```diff
@@ -28,225 +28,268 @@ import (
...
+	return fieldsInv, nil
+}
+
+func docValueCmd(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("must specify index file path")
+	}
+
+	data := segment.Data()
+	log.Printf("hello\n")
```
metonymic-smokey, line 218: Remove this line.

abhinavdangeti, line 218: Drop log

cmd/zap/cmd/fields.go
```diff
@@ -28,33 +27,51 @@ var fieldsCmd = &cobra.Command{
...
-			n += uint64(read)
-
-			var nameLen uint64
-			nameLen, read = binary.Uvarint(data[addr+n : fieldsIndexEnd])
-			n += uint64(read)
+		var fieldInv []string
+		pos := segment.FieldsIndexOffset()
+		if pos == 0 {
+			// this is the case only for older file formats
+			return fmt.Errorf("file format not supported?")
```
abhinavdangeti, line 41: lose the `?`

abhinavdangeti (changes requested):
@Thejas-bhat it'd be ideal if you could re-use methods rather than re-implement them again for the command line tooling. It'd be double work for us to maintain this otherwise. Is there any code here that's common with access methods elsewhere?

cmd/zap/cmd/vector.go
```diff
@@ -0,0 +1,171 @@
...
+				case 2:
+					metrics := map[int]string{
+						faiss.MetricInnerProduct:  "inner product",
+						faiss.MetricL2:            "l2 distance",
+						faiss.MetricL1:            "manhattan distance",
+						faiss.MetricLinf:          "chebyshev distance",
+						faiss.MetricLp:            "lp distance",
+						faiss.MetricCanberra:      "canberra distance",
+						faiss.MetricBrayCurtis:    "bray curtis distance",
+						faiss.MetricJensenShannon: "jensen shannon distance",
```
abhinavdangeti, line 85: We don't support any of these other than first 2, so no need to accommodate these.

cmd/zap/cmd/root.go
```diff
@@ -24,6 +24,11 @@ import (
 
 var segment *zap.Segment
 
+const (
+	sectionInvertedTextIndex = iota
+	sectionFaissVectorIndex
+)
```
abhinavdangeti, line 30: We don't need to redeclare these here, use zap.sectionInvertedTextIndex and zap.sectionFaissVectorIndex

abhinavdangeti (changes requested):
@Thejas-bhat any thing else that needs look into in light of my earlier comment?

cmd/zap/cmd/root.go
Thejas-bhat: so i'd have to export those two vars in section.go. i've made the change, please let me know that's fine.

Thejas-bhat:
> @Thejas-bhat any thing else that needs look into in light of my earlier comment?

i've reused a bunch of functions, however the diff shows a lot of change mainly because i moved the content of the docvalue command to a separate function.

cmd/zap/cmd/root.go
abhinavdangeti: Ya export away.

abhinavdangeti (changes requested):

cmd/zap/cmd/footer.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 // footerCmd represents the footer command
+// - support the sectionsIndexOffset over here?
```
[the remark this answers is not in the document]
Thejas-bhat: since the fieldsIndexOffset and sectionsIndexOffset both point to the same info, I think its better to retain the existing FieldsIndexOffset and not introduce a separate sectionsIndexOffset. Do you think that's a good idea?

cmd/zap/cmd/explore.go
```diff
@@ -28,6 +28,7 @@ import (
 const termNotEncoded = 0
 
 // exploreCmd represents the explore command
+// no big changes over here as well.
```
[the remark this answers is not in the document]
Thejas-bhat: done.

cmd/zap/cmd/vector.go
Thejas-bhat: sorry about that, i think i forgot to save the files before the push :p

cmd/zap/cmd/footer.go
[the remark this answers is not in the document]
abhinavdangeti: If the user is going to remain oblivious of this stuff, then I think it's fine.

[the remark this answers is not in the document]
Thejas-bhat: i dont think it should affect the user in my opinion, since we'd not be advertising those two offsets (which are equal) separately and i dont think its necessary to expose it to user either actually.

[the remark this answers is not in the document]
abhinavdangeti: Ok, let's go with this for now, let's do our due diligence in testing this as well.

abhinavdangeti (approved):

PR #180, opened 2023-11-14
replacing the flat index with the IVF family index.

Thejas-bhat:


abhinavdangeti (approved):

section_faiss_vector_index.go
```diff
@@ -398,16 +378,22 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
 			metric = faiss.MetricInnerProduct
 		}
 
-		// create an index, its always a flat for now, because each batch size
-		// won't have too many vectors (in order for >100K). todo: will need to revisit
-		// this logic - creating based on configured batch size in scorch.
-		index, err := faiss.IndexFactory(int(content.dim), "IDMap2,Flat", metric)
+		// create an index - currently its a IVF with 2 clusters and no compression
+		index, err := faiss.IndexFactory(int(content.dim), "IVF2,Flat", metric)
```
abhinavdangeti, line 382: This means there's a requirement that the index needs at least 2 vectors? So if a batch comes in with a single vector - this'd fail.

section_faiss_vector_index.go
```diff
@@ -271,85 +271,64 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(fieldID int, sbs []*Segme
...
+	metric := vecIndexes[0].MetricType()
+	finalVecIDs := maps.Keys(vecToDocID)
+
+	var indexType string
+	// todo: perhaps the index type can be chosen from a config and tuned accordingly.
+	switch {
+	case len(vecToDocID) > 10000:
+		indexType = "IVF100,SQ8"
+	default:
+		indexType = "IVF10,Flat"
```
abhinavdangeti, line 301: Putting in a requirement of at least 10 vectors to survive the merge is not a good idea.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
abhinavdangeti: So should probably use IVF1.

abhinavdangeti:
@Thejas-bhat I'm thinking you should introduce a common helper for the introduce path and merge path where ..

```
if vectorCount >= 100 {
    indexType = "IVF100,SQ8"
} else if vectorCount >= 10 {
    indexType = "IVF10,Flat"
} else {
    indexType = "IVF1,Flat"
}
```

abhinavdangeti (approved):

PR #181, opened 2023-11-14
optimization: avoiding storing of bitmaps when vector is in a single document

Thejas-bhat:
- the aim of the PR is to avoid storing a bitmap when the vector is present in a single document. this helps to avoid building the bitmap from a buffer and just do a cheaper decoding the uint value from the bytes on the index file.

### DIFF
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -309,16 +309,25 @@ func (sb *SegmentBase) SimilarVectors(field string, qVector []float32, k int64, 
 				vecID, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
 				pos += n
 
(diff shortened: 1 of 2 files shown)


faiss_vector_posting.go
```diff
@@ -309,19 +309,29 @@ func (sb *SegmentBase) SimilarVectors(field string, qVector []float32, k int64,
...
 				pos += n
 
-				roaringBytes := sb.mem[pos : pos+int(bitMapLen)]
-				pos += int(bitMapLen)
-
 				bitMap := roaring.NewBitmap()
-				_, err := bitMap.FromBuffer(roaringBytes)
-				if err != nil {
-					return nil, err
+				if numDocs > 1 {
```
abhinavdangeti, line 316: Let's add some in-line commentary around this condition - why  > 1?

abhinavdangeti (approved):

PR #182, opened 2023-11-16
MB-59692: Respect choice of similarity metric, bleve_index_api@v1.1.3

abhinavdangeti:


section_faiss_vector_index.go
```diff
@@ -368,10 +368,15 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
 			ids = append(ids, int64(hash))
 		}
 
+		var metric = faiss.MetricL2
+		if content.metric == index.CosineSimilarity {
```
Thejas-bhat, line 372: i feel like it would be a lot cleaner if we declare the map in the bleve_index_api domain as:

```
var SupportedSimilarityMetrics = map[string]int{
	EuclideanDistance: faiss.MetricL2,
	CosineSimilarity:  faiss.MetricInnerProduct,
}
```
over here what we could do is pass the index.SupportedSimilarityMetrics[metric] value to the factory method

abhinavdangeti: Ya I thought about doing that earlier, but perhaps is not a good idea to pollute the generic bleve_index_api with implementation specifics - referring to go-faiss here, in case we decide to embed a different better performing vector index in the future.

Thejas-bhat (approved):

PR #184, opened 2023-11-21
Separate functions to read and search vector indexes.

metonymic-smokey:


faiss_vector_posting.go
```diff
@@ -266,7 +267,44 @@ func (vpl *VecPostingsIterator) BytesWritten() uint64 {
...
+			if err != nil {
+				return vecIndex, err
+			}
+		}
+	}
+
+	return vecIndex, nil
+}
+
+func (sb *SegmentBase) SearchSimilarVectors(vecIndex index.VectorIndex, field string, qVector []float32, k int64,
```
abhinavdangeti, line 305: This API definition doesn't feel good - where it's a member of `SegmentBase` but will only read content from vecIndex which was obtained previously from `sb`. But i understand why you need it this way - so given that it's an exported API, let's add a check here `if vecIndex == nil` - in which case we call GetVectorIndex(..) above.

Also, more importantly - it seems we're repeating a bunch of things we already do in GetVectorIndex above below - which needs some consolidation?

abhinavdangeti (changes requested):

faiss_vector_posting.go
metonymic-smokey: I forgot to bring this up during the dev call but I pushed a fix for making this a common func(point 2).
Pls let me know what you think.

faiss_vector_posting.go
```diff
@@ -266,7 +266,60 @@ func (vpl *VecPostingsIterator) BytesWritten() uint64 {
 	return 0
 }
 
-func (sb *SegmentBase) SimilarVectors(field string, qVector []float32, k int64, except *roaring.Bitmap) (segment.VecPostingsList, error) {
+func (sb *SegmentBase) GetVectorIndex(field string) (segment.VectorIndex, error) {
```
abhinavdangeti, line 269: Add a comment above here, that the caller is now responsible for closing the `vecIndex`.
Will add this in separate PR.

abhinavdangeti (approved):

PR #185, opened 2023-11-22
MB-59447: support nested/chunked vectors

moshaad7:
(*vectorIndexOpaque)process can now handle vectors with length greater than dims ( but multiple of dims).

It will keep reading dims number of floats and index a vector formed out of them.

Related bleve change: https://github.com/blevesearch/bleve/pull/1909


section_faiss_vector_index.go
```diff
@@ -510,32 +510,41 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
...
 			}
 		}
 		// add the docID to the bitmap
-		vo.vecIDMap[vecHash].docIDs.Add(docNum)
+		vo.vecIDMap[subVecHash].docIDs.Add(docNum)
 
 		// tracking the unique vectors for every field which will be used later
 		// to construct the vector index.
 		if _, ok := vo.vecFieldMap[fieldID]; !ok {
 			vo.vecFieldMap[fieldID] = indexContent{
```
moshaad7, line 536: Optimization:

vecIDMap can be restructured to store only slice of hashValues.
Instead of a map of <hashValue, vecInfo>

faiss_vector_test.go
```diff
@@ -135,34 +134,32 @@ func newStubFieldVec(name string, vector []float32, d int, metric string, fieldO
...
-	}
-	return rv
+var stubVecData = [][]float32{
+	{1.0, 2.0, 3.0},
+	{12.0, 42.6, 78.65},
+	{6.7, 0.876, 9.45},
+	{7.437, 9.994, 0.407},
+	{4.439, 0.307, 1.063},
+	{6.653, 7.752, 0.972},
+	{1.23, 2.45, 2.867, 3.33, 4.56, 5.67},
```
moshaad7, line 144: add comment here

faiss_vector_test.go
```diff
@@ -390,13 +435,17 @@ func TestVectorSegment(t *testing.T) {
 		t.Fatalf("expected %d vecs got %d vecs", vecIndex.Ntotal(), numVecs)
 	}
 
+	queryVec := []float32{0.0, 0.0, 0.0}
+	hitDocIDs := []uint64{2, 9, 9}
```
moshaad7, line 439: bleve should deduplicate these.
Add a unit test there

section_faiss_vector_index.go
abhinavdangeti: @moshaad7 Both this and the earlier optimization you cite sound good to accommodate within these PRs.

section_faiss_vector_index.go
```diff
@@ -564,32 +564,41 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
 	dim := field.Dims()
 	metric := field.Similarity()
 
-	if vec != nil {
+	// Embedder is supposed to make sure len(vec) is a multiple of dim.
```
abhinavdangeti, line 567: Embedder? :)

Use a simpler word - caller?

section_faiss_vector_index.go
```diff
@@ -564,32 +564,41 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
...
+	// Not double checking it here to avoid the overhead.
+	numSubVecs := len(vec) / int(dim)
+	for i := 0; i < numSubVecs; i++ {
+		subVec := vec[i*int(dim) : (i+1)*int(dim)]
```
abhinavdangeti, line 571: Are the `int` casts here necessary?

section_faiss_vector_index.go
```diff
@@ -564,32 +564,41 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
...
+
+		if subVec == nil {
```
abhinavdangeti, line 573: When can this be `nil`? And why continue if it is?

section_faiss_vector_index.go
```diff
@@ -510,32 +510,41 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
...
+			continue
+		}
+
 		// NOTE: currently, indexing only unique vectors.
-		vecHash := hashCode(vec)
-		if _, ok := vo.vecIDMap[vecHash]; !ok {
-			vo.vecIDMap[vecHash] = vecInfo{
+		subVecHash := hashCode(subVec)
```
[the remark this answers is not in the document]
abhinavdangeti: Wouldn't this mean that you have pass the vecHashs down from bleve to zap?

section_faiss_vector_index.go
```diff
@@ -564,32 +564,41 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
 	dim := field.Dims()
 	metric := field.Similarity()
 
-	if vec != nil {
```
abhinavdangeti, line 567: I don't think we should drop this `if vec != nil` to avoid any out of bounds reference below.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
```diff
@@ -564,32 +564,41 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
...
+	// Embedder is supposed to make sure len(vec) is a multiple of dim.
+	// Not double checking it here to avoid the overhead.
+	numSubVecs := len(vec) / int(dim)
+	for i := 0; i < numSubVecs; i++ {
```
Thejas-bhat, line 570: now i don't have any benchmark numbers of something to suport this doubt, however in a purely non-nested situation where the loop is iterated only once, could there be a perf impact due to doing the following code as part of for loop iteration rather than just doing it as part of a conditional check of `if len(vec) == int(dim)`. 

i dont know too much about the golang internals, but yeah just a doubt i had when i glanced through the code, so just a small food for thought ig

section_faiss_vector_index.go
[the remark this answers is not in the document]
moshaad7: Yes, vecHashs will have to passed down to zap from bleve. 
That's the price we will have to pay, memory for latency. Only during the existence of interim, we will keep the hash in memory.

I will continue this discussion in later PRs.

section_faiss_vector_index.go
moshaad7: Sorry I wanted to say something else, but typed other thing.

We have two maps and redundant content in them

```

vecIDMap map[int64]vecInfo
vecFieldMap map[uint16]indexContent

type vecInfo struct {
	vec    []float32
	docIDs *roaring.Bitmap
}

type indexContent struct {
	vecs   map[int64]vecInfo
	dim    uint16
	metric string
}
```

You can see that vecInfo is redundant in both indexContent and vecIDMap. 

So, indexContent should only store vector hashes, and if it want to know the vecInfo, it should consult vecIDMap

Will continue this conversation in new PR

section_faiss_vector_index.go
moshaad7: done

section_faiss_vector_index.go
moshaad7: done

section_faiss_vector_index.go
moshaad7: no, removed them

section_faiss_vector_index.go
moshaad7: We don't need to perform nil vec checks here, as bleve will already take care of it.

section_faiss_vector_index.go
moshaad7: Thanks. 
Will continue this discussion in the next PR ( targeting some optimizations)

abhinavdangeti (approved):

PR #187, opened 2023-11-23
bug fix: fixing the logic of tracking vector ids during a merge

Thejas-bhat:
- when there is a doc update, there could be a case where the number of valid docs for a vector (after remapping to the new docIDs) is zero. in this case, we shouldn't account the vector to be included in reconstruction and also should be considered as something invalid (deleted) so that we can remove that vector from the index.
- fixing the decoding of segment data in `vector` cmd tool.
- fixing the vector code formation on the postings list. essentially the score can be negative in case of dot product, however due to the right shift being 31 earlier we were reading the sign bit to be the part of docNum as well which caused incorrect doc nums to be returned. the fix for this is to shift the bits by right the number = 32.

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -158,6 +158,10 @@ LOOP:
 				// of that segment in the resulting bitmap
 				bitMap = remapDocIDs(bitMap, newDocNumsIn[segI])
 				if vecToDocID[vecID] == nil {
+					// if the remapped bitmap has valid docs as entries, track it
+					// as part of vecs to be reconstructed (for larger indexes).
+					// otherwise, since there are no docs present for this vecID,
+					// delete it from the specific vector index later on.
 					if bitMap.GetCardinality() > 0 {
 						vecToDocID[vecID] = bitMap
 						indexes[len(indexes)-1].vecIds = append(indexes[len(indexes)-1].vecIds, vecID)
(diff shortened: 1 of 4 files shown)


section_faiss_vector_index.go
```diff
@@ -153,18 +154,16 @@ LOOP:
...
 				if vecToDocID[vecID] == nil {
-					// if there are some tombstone entries in the docIDs, marked
-					// in the drops[ith-segment] bitmap, don't include them in the
-					// final bitmap.
-					if drops[segI] != nil && !drops[segI].IsEmpty() {
-						vecToDocID[vecID] = roaring.AndNot(bitMap, drops[segI])
-					} else {
+					if bitMap.GetCardinality() > 0 {
 						vecToDocID[vecID] = bitMap
+						indexes[len(indexes)-1].vecIds = append(indexes[len(indexes)-1].vecIds, vecID)
```
abhinavdangeti, line 163: I'd advise some commentary here on why the particular index of the `indexes` array.

section_faiss_vector_index.go
```diff
@@ -252,6 +251,19 @@ func (v *vectorIndexOpaque) flushVectorSection(vecToDocID map[int64]*roaring.Bit
 	return fieldStart, nil
 }
 
+func removeInvalidVecs(index *faiss.IndexImpl, ids []int64) error {
```
abhinavdangeti, line 254: Refactor to `removeDeletedVectors` to more accurately reflect what this is for.

abhinavdangeti (changes requested):

abhinavdangeti (approved):

PR #188, opened 2023-11-27
fixing IO stats computation

Thejas-bhat:
- includes fixes related to the io stats computation (bytesWritten and bytesRead) in light of new index sections PR.


section_inverted_text_index.go
```diff
@@ -911,8 +931,9 @@ type invertedIndexOpaque struct {
 
 	fieldAddrs map[int]int
 
-	fieldsSame bool
```
abhinavdangeti, line 914: I know this was always there - but what a bad attribute name :)

abhinavdangeti (approved):

PR #192, opened 2023-12-07
cache eviction policy for vector indexes 

Thejas-bhat:
- Uses the exponentially weighted moving average to get the average hits on a particular field (thereby a particular vector index). The index stays in the cache above a certain threshold of the average hits and below which the index is closed and the memory is freed for reuse on C side of things. 
- This ensures that we are not keeping the index in the cache (and holding up memory) even when there is no query workload on the field in a segment so the memory pressure does get reduced on the C side of operations.
- Still a work in progress


Thejas-bhat:
Closing this in favour of https://github.com/blevesearch/zapx/pull/226

PR #195, opened 2023-12-11
MB-60068: Case when dimensionality of index and query vector mismatches

abhinavdangeti:


abhinavdangeti:
Let's not do this here, it's better positioned in go-faiss.

abhinavdangeti:
Keeping this here, because it seems we do the validation for this either in bleve (index time) and not in go-faiss. So let's retain the validation in bleve/zapx for now.

PR #197, opened 2023-12-12
MB-60084: Change index merge procedure to account for existing index type.

metonymic-smokey:
Consider the following case:
A segment with an existing IVF index has a large number of deleted vectors due to mutations, etc. Post merge, due to the lesser number of vectors, a flat index will be created. In the current approach, we were attempting to remove vector IDs from an IVF index with a hashtable using a batch ID selector which isn't supported. Hence, the panic seen in the ticket.

This PR adopts an approach where mergeFrom() is applied only if the existing indexes are all flat and the new index to be formed is flat too.

In the process, I've also made some minor changes:
1.  limited training to only IVF indexes - flat indexes don't need to be trained.
2. done a minor refactoring of the utilities used by related UTs.

### DIFF
--- a/faiss_vector_test.go
+++ b/faiss_vector_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math"
+	"math/rand"
 	"os"
 	"testing"
 
@@ -134,6 +135,14 @@ func newStubFieldVec(name string, vector []float32, d int, metric string, fieldO
 		encodedType: 'v',
 		options:     fieldOptions,
 	}
+}
+
+func stubVecDataLengthN(n int) [][]float32 {
+	rv := make([][]float32, n)
+	for i := 0; i < n; i++ {
+		rv[i] = []float32{rand.Float32(), rand.Float32(), rand.Float32()}
+	}
+	return rv
 }
 
 func stubVecData() [][]float32 {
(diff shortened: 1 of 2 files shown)


faiss_vector_test.go
```diff
@@ -335,6 +344,43 @@ func letsCreateVectorIndexForTesting(dataset [][]float32, dims int, similarity s
...
 }
 
+func TestRemoveVectorIDs(t *testing.T) {
+
+	testCases := []struct {
+		numVecs int
+	}{
+		{
+			// 5 vectors to test for flat indexes.
+			numVecs: 5,
```
metonymic-smokey, line 354: TODO Find a way to ensure all the types of vec indexes are tested in such UTs without using hard limits like 5/500 which may be prone to staleness if the code is updated.

section_faiss_vector_index.go
```diff
@@ -256,7 +256,7 @@ func (v *vectorIndexOpaque) flushVectorSection(vecToDocID map[int64]*roaring.Bit
 }
 
 func removeDeletedVectors(index *faiss.IndexImpl, ids []int64) error {
-	sel, err := faiss.NewIDSelectorBatch(ids)
+	sel, err := faiss.NewIDSelectorArray(ids)
```
metonymic-smokey, line 259: TODO(maybe for a future PR): should the selector be deleted to free memory, like `sel.Delete()`?

abhinavdangeti:
@Thejas-bhat would you review this please?

Thejas-bhat:
actually i'm a bit confused over here. the thing is, the func getIndexType() its based on the number of vectors post-merge that are going to be there. if the post-merge number of vectors is greater than 390 it'll switch to IVF - now it's true that this means the participating child indexes can be either flat or IVF, however the final index is going to be IVF and the flag returned is true. if we see the block of code following the getIndexType() we see that the removeIDs stuff is done only if the flag is false - the final index is flat. this also means if that block is hit, none of the constituent indexes is IVF given the logic of getIndexType(). 

so, i'm unsure why we'd be trying to remove vectors from an IVF index in that else condition. also, when it comes to IVF type of indexes since we anyways have to reconstruct the vectors - only the valid ones though, so the deleted ones aren't reconstructed - the creation of an IDSelector and then doing a removeIDs on it wouldn't be necessary since the reconstruction of vectors phase is handling it appropriately. 

Aditi, could you please explain me with an example as to in which scenario we're trying to removeIDs from an IVF index? like i might be missing something over here.

metonymic-smokey:
So I'm aware of how index types are chosen during merge :)
I added that case since this error happens only with IVF indexes where we set DirectMap to a hashtable so it must have been an IVF index that is getting deleted.
If it was a flat map, we'd have seen a different error if we were indeed using a direct map(which we aren't) since remove IDs isn't supported with direct maps - https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVFFlat.cpp#L430-L431

metonymic-smokey:
While the type of vector to be formed post merge is determined by the count of the updated vectors, the existing index types may not be suitable for direct merging/removing IDs.
1. In a situation where segments with IVF indexes, each with a high number of deletes, needs to be combined into a flat index, they need to be reconstructed and the flat index constructed _also taking into account the existing index type_. Currently, we're doing a mergeFrom() which would merge the two without re-training.

2. If a situation arises where vector indexes of 2 different types ended up being merged using a mergeFrom(), the merge operation would error since they're incompatible for merging as is done currently.

section_faiss_vector_index.go
metonymic-smokey: Done in the current PR itself.

section_faiss_vector_index.go
```diff
@@ -255,6 +255,26 @@ func (v *vectorIndexOpaque) flushVectorSection(vecToDocID map[int64]*roaring.Bit
...
+// flat index.
+func reconstructionRequired(isNewIndexIVF bool, indexes []vecIndexMeta) bool {
+	// Always required for IVF indexes.
+	if isNewIndexIVF {
+		return true
+	}
+
+	// if any existing index is not flat, all need to be reconstructed.
+	for _, index := range indexes {
+		_, isExistingIndexIVF := getIndexType(len(index.vecIds) + len(index.deleted))
```
metonymic-smokey, line 269: TODO Food for thought: Would this cause an issue with forward compatibility if the limits deciding flat/IVF are changed in a newer zapx version the user upgrades to?

abhinavdangeti (approved):
Looks good, pending the one minor naming comment.

abhinavdangeti (approved):

PR #199, opened 2023-12-21
MB-60152: Panic fix while trying to fetch metadata of a vector index

Thejas-bhat:
- The `mergeAndWriteVectorIndexes() ` is called whenever we are trying to merge multiple vector indexes. As part of that we are tracking the participating vector indexes' metadata as part of `vecIndexMeta` - which has the ids corresponding to the set of valid vectors and also the ids of those vectors that have been deleted/are obsolete. 
- Earlier, we were passing all the segments to the func and merging the valid vector indexes. However, that caused issues with respect to accessing nil/empty values. 
- The PR corrects this by passing only the segments having the valid vector content in it.

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -66,11 +66,6 @@ type vecIndexMeta struct {
 	deleted     []int64
 }
 
-type vecIndexData struct {
-	index *faiss.IndexImpl
-	meta  *vecIndexMeta
-}
-
 func remapDocIDs(oldIDs *roaring.Bitmap, newIDs []uint64) *roaring.Bitmap {
 	newBitmap := roaring.NewBitmap()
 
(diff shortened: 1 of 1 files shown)


moshaad7:
Let's not call it refactoring if this fix is addressing a panic

section_faiss_vector_index.go
```diff
@@ -322,20 +323,32 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(fieldID int, sbs []*Segme
 
 	if reconstructionRequired(isIVF, indexes) {
 		// merging of indexes with reconstruction
-		// method. the indexes[i].vecIds is such that it has only the valid vecs
+		// method. the meta.indexes[i].vecIds is such that it has only the valid vecs
```
moshaad7, line 326: Is this intentional

Thejas-bhat: my bad, i had to clean it up

moshaad7 (approved):

metonymic-smokey (approved):

PR #202, opened 2024-01-04
MB-60269 - Merge path fixes

metonymic-smokey:
This PR addresses 2 merge path issues - 

A. IDs need to be added to a vector index in the _same order as the vectors_ to maintain the integrity of the mapping between vectors and IDs which is important for both search and reconstruction. 
When reconstructing indexes, the current approach uses the keys of a map to decide the order of the final vector IDs. However, since the Keys() function provides no guarantee of ordering of the returned list, the vectors and IDs are not guaranteed to be inserted in the same order.

B. Currently, merging flat indexes involves adding all existing flat indexes into an existing index. 
1. This indexes duplicate vectors - since a hash is unique to a vector, 2 segments with the same vectors in an index will have the same hash. merge_from() adds duplicate vectors to the index.
2. This does not take into account updated doc IDs since those aren't updated in the constituent vector indexes being merged.
This PR uses the reconstruction method for flat indexes too and rebuilds the flat index with updated data and IDs.

Testing - 
Performed local testing with the SIFT Small dataset(10k vectors) with k = 2/3 for the following cases:
1. recall - saw an increase in the number of common results with FTS and FAISS groundtruth vectors.
2. results before and after a rebuild - remained the same.
3. results when cloning an index - remained the same for both.

Picked the dataset since this has a reasonable recall with the current index configuration.

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -258,27 +258,6 @@ func (v *vectorIndexOpaque) flushVectorSection(vecToDocID map[int64]*roaring.Bit
 	return fieldStart, nil
 }
 
-// Func which returns if reconstruction is required.
-// Won't be required if multiple flat indexes are being combined into a larger
-// flat index.
-func reconstructionRequired(isNewIndexIVF bool, indexes []*vecIndexMeta) bool {
-	// Always required for IVF indexes.
-	if isNewIndexIVF {
-		return true
-	}
-
-	// if any existing index is not flat, all need to be reconstructed.
-	for _, index := range indexes {
-		_, isExistingIndexIVF := getIndexType(len(index.vecIds) +
-			len(index.deleted))
-		if isExistingIndexIVF {
-			return true
-		}
-	}
-
-	return false
-}
-
 func removeDeletedVectors(index *faiss.IndexImpl, ids []int64) error {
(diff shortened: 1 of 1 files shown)


metonymic-smokey:
TODO Open go-faiss PR to clean up MergeFrom() now?

section_faiss_vector_index.go
```diff
@@ -321,105 +285,78 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(fieldID int, sbs []*Segme
...
-		finalVecIDs := maps.Keys(vecToDocID)
+	var finalVecIDs []int64
 
-		index, err := faiss.IndexFactory(dims, indexType, metric)
-		if err != nil {
-			return err
+	// merging of indexes with reconstruction
+	// method. the indexes[i].vecIds is such that it has only the valid vecs
+	// of this vector index present in it, so we'd be reconstructed only the
+	// valid ones.
```
abhinavdangeti, line 293: Fix formatting the commentary above.

abhinavdangeti (commented):
@metonymic-smokey like we spoke, would you add some backing tests or results at the least on how things looked with this change.

@blevesearch/collaborators must take another look.

Thejas-bhat (approved):

abhinavdangeti (approved):

PR #203, opened 2024-01-09
Changing the defaults for centroid count and nprobe.

metonymic-smokey:


section_faiss_vector_index.go
```diff
@@ -382,15 +388,40 @@ func (v *vectorIndexOpaque) grabBuf(size int) []byte {
...
+		// 100 points per cluster is a reasonable default, considering the default
+		// minimum and maximum points per cluster is 39 and 256 respectively.
+		// Since it's a recommendation to have a minimum of 10 clusters, 1000(100 * 10)
+		// was chosen as the lower threshold.
+		nlist = nVecs / 100
+	}
+	return nlist
+}
+
+const IndexTypeIVF = 1
```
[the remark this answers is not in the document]
metonymic-smokey: done, pls take a look.

section_faiss_vector_index.go
```diff
@@ -382,15 +388,42 @@ func (v *vectorIndexOpaque) grabBuf(size int) []byte {
...
+// and an index type.
+func getIndexType(nVecs, nlist int) (string, int) {
 	switch {
-	case nVecs > 10000:
-		return "IVF100,SQ8", true
-	// default minimum number of training points per cluster is 39
-	case nVecs >= 390:
-		return "IVF10,Flat", true
+	case nVecs >= 10000:
+		return "IVF" + strconv.Itoa(nlist) + ",SQ8", IndexTypeIVF
```
abhinavdangeti, line 421: Better move would be to

```
return fmt.Sprintf("IVF%d,SQ8", nlist)
``` 

.. rather than declaring 3 strings and concats. Same below.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
metonymic-smokey: yeah, my bad, good catch.

section_faiss_vector_index.go
```diff
@@ -413,21 +445,24 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
 			metric = faiss.MetricInnerProduct
 		}
 
-		indexType, isIVF := getIndexType(len(ids))
+		nlist := getNumCentroids(len(ids))
+		indexDescription, indexClass := getIndexType(len(ids), nlist)
```
abhinavdangeti, line 449: Much better, let's do a little more refactoring here ..

```
indexDescription, indexClass := determineIndexToUse(len(ids))
```

Now the definition would be this instead of getIndexType above -
```
func determineIndexToUse(nVecs int) (string, int) {
        var nlist int
	switch {
	// At 1M vectors, nlist = 4k gave a reasonably high recall with the right nprobe,
	// whereas 1M/100 = 10000 centroids would increase training time without
	// corresponding increase in recall
	case nVecs >= 1000000:
		nlist = int(4 * math.Sqrt(float64(nVecs)))
	case nVecs >= 1000:
		// 100 points per cluster is a reasonable default, considering the default
		// minimum and maximum points per cluster is 39 and 256 respectively.
		// Since it's a recommendation to have a minimum of 10 clusters, 1000(100 * 10)
		// was chosen as the lower threshold.
		nlist = nVecs / 100
	}

	switch {
	case nVecs >= 10000:
		return fmt.Sprintf("IVF%d,SQ8", nlist), IndexTypeIVF
	case nVecs >= 1000:
		return fmt.Sprintf("IVF%d,Flat", nlist), IndexTypeIVF
	default:
		return "IDMap2,Flat", IndexTypeFlat
	}
}
```

Drop the separate getNumCentroids api.

abhinavdangeti: Remember it is in our interest to reduce the number of ops we do along these hot code paths - even something as simple as `len(array)`.

metonymic-smokey: Re dropping the getNumCentroids() func, I retained it as a separate API since the nlist is going to be used when calculating nprobe only for IVF indexes.
An alternative would be to refactor as you said and introduce a new go-faiss API to get the number of centroids for an IVF index - I avoided this since I think using a local var is cheaper than a call to faiss.GetNlist() or some similar FAISS API. Again, considering this is a hot path, it made sense to me to avoid FAISS calls(go -> cgo -> C and back) for a value that was already computed.

metonymic-smokey: I am changing the function name though since that's better naming.

metonymic-smokey: Pls resolve if this makes sense, thanks

PR #204, opened 2024-01-09
MB-60029 - Persisting index tunables

metonymic-smokey:
This PR -
1. accommodates user choice of optimising for speed/recall when creating an index - it does this by changing the nprobe count based on the index type.
3. persists this so that this is adhered to during the merge process as well.


metonymic-smokey:
TODO: Fix failing UT.

abhinavdangeti:
I think I want a better understanding of why we're trying to persist this. I feel we need to first resolve what the config is going to look like.

section_faiss_vector_index.go
```diff
@@ -76,6 +79,10 @@ func remapDocIDs(oldIDs *roaring.Bitmap, newIDs []uint64) *roaring.Bitmap {
 	return newBitmap
 }
 
+// constants for the purpose of persisting the index type as an int
+const recallIndexInt = 0
+const latencyIndexInt = 1
```
abhinavdangeti, line 84: This is very brittle, declaring different constants here. You should at least get these numbers for recall or latency by setting up values for the map you've declared here -  https://github.com/blevesearch/bleve_index_api/pull/46 .

section_faiss_vector_index.go
```diff
@@ -114,13 +121,21 @@ func (v *faissVectorIndexSection) Merge(opaque map[int]resetable, segments []*Se
...
 
+			indexTypeInt, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+			pos += n
+
 			indexSize, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
 			pos += n
 
 			vecSegs = append(vecSegs, sb)
+			indexOptimizationStr := index.IndexOptimizationForRecall
+			if indexTypeInt == latencyIndexInt {
```
abhinavdangeti, line 132: This is NOT an indexType.

section_faiss_vector_index.go
```diff
@@ -382,15 +419,41 @@ func (v *vectorIndexOpaque) grabBuf(size int) []byte {
 	return buf[0:size]
 }
 
-func getIndexType(nVecs int) (string, bool) {
+// Return the number of centroids for an IVF index.
+func getNumCentroids(nVecs int) int {
```
abhinavdangeti, line 423: Wonder why this code is showing up here, perhaps this commits needs a rebase now that the base branch has moved from centroids2 to master.

abhinavdangeti (changes requested):

abhinavdangeti:
@metonymic-smokey I rebased - but unfortunately I had to force merge to avoid including the parent commits as part of this PR. Please checkout another local `centroids3` branch based on origin/centroids3 (after a `git remote update`) before you push up more commits here.

faiss_vector_test.go
```diff
@@ -426,6 +431,7 @@ func TestVectorSegment(t *testing.T) {
 		numVecs, _ := getSectionContentOffsets(vecSegBase, stubVecFieldStartAddr)
 
 	if docValueStart != fieldNotUninverted {
+		fmt.Printf("doc value start: %v \n", docValueStart)
```
abhinavdangeti, line 434: Drop the debug print statement.

cmd/zap/cmd/vector.go
```diff
@@ -128,6 +128,9 @@ func decodeSection(data []byte, start uint64) (int, int, map[int64][]uint32, *fa
 	_, n = binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
 	pos += n
 
+	_, n = binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
+	pos += n
+
```
abhinavdangeti, line 133: I realize this code duplication wasn't introduced here, but this PR is sure building on it:
Lines 125 through 133 show so much repetition. For starters -
```
var n int
for i := 0 ; i < 3; i++ {
    _, n = binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
    pos += n
}
```

Also the comment above needs fixing and additional commentary around why this needs to occur 3 times.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
```diff
@@ -63,10 +63,17 @@ func (v *faissVectorIndexSection) AddrForField(opaque map[int]resetable, fieldID
...
-	deleted     []int64
+	startOffset       int
+	indexSize         uint64
+	vecIds            []int64
+	deleted           []int64
+	indexOptimizedFor string
+}
+
+// maps the index optimization type int to the string
+var indexOptimizationMap = map[int]string{
```
abhinavdangeti, line 74: Not a good idea to add this reverse lookup here, let's keep it all in one place - within bleve_index_api.

section_faiss_vector_index.go
```diff
@@ -216,6 +234,16 @@ func (v *vectorIndexOpaque) flushVectorSection(vecToDocID map[int64]uint64,
 	return fieldStart, nil
 }
 
+// Calculates the nprobe count, given nlist(number of centroids) based on
+// the metric the index is optimized for.
+func calculateNprobe(nlist int, indexOptimizedFor string) int32 {
+	nprobe := int32(math.Sqrt(float64(nlist)))
+	if indexOptimizedFor == index.IndexOptimizedForLatency {
+		nprobe /= 2
```
abhinavdangeti, line 242: Must handle the case where nprobe is less than 2 - in which case here we'd be setting it to zero.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
metonymic-smokey: The thresholds in determineIndexType() were chosen to ensure that there are never less than 10 centroids(1000/100) so the minimum values of nprobe would be around 3.
Still added a check to be on the safe side but just something to keep in mind.

cmd/zap/cmd/vector.go
```diff
@@ -119,13 +119,14 @@ func decodeSection(data []byte, start uint64) (int, int, map[int64]uint64, *fais
...
-	// loading doc values - adhering to the sections format. never
+	// the below loop loads the following:
+	// 1. doc values - adhering to the sections format. never
 	// valid values for vector section
-	_, n := binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
-	pos += n
-
-	_, n = binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
-	pos += n
+	// 2. index optimization type.
```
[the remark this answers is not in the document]
metonymic-smokey: The first 2 iterations are used to load doc values, the third for the index optimisation type. 
Updated the comment to reflect it.

abhinavdangeti (changes requested):
@metonymic-smokey We've a unit test failure with this that you seen to still not have addressed - 
```
--- FAIL: TestVectorSegment (0.01s)
    faiss_vector_test.go:430: vector field doesn't support doc values
```

Please fix this and raise a patch here. Remember to `git pull` before raising a commit and **no** force pushes please.

abhinavdangeti (approved):

PR #205, opened 2024-01-11
MB-60202 - Filtering deleted documents during FAISS vector index search

metonymic-smokey:
This PR addresses filtering of deleted documents by moving from post-filtering to search time filtering. 

- Passes doc IDs to exclude for each search.
- To account for duplicate vectors: saves every vector with a unique hash so that the selector only excludes the specific vector ID. The current approach of identical vectors having identical hashes would lead to (unnecessarily) excluding vectors if other (identical) vectors in the segment were deleted.

### DIFF
--- a/cmd/zap/cmd/vector.go
+++ b/cmd/zap/cmd/vector.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"strconv"
 
-	"github.com/RoaringBitmap/roaring"
 	"github.com/blevesearch/go-faiss"
 	zap "github.com/blevesearch/zapx/v16"
 	"github.com/spf13/cobra"
@@ -91,8 +90,8 @@ var vectorCmd = &cobra.Command{
 				case 3:
 					if args[2] == "list" {
 						fmt.Printf("listing the vector IDs in the index\n")
-						for vecID, docs := range vecDocIDMap {
-							fmt.Printf("vector with vecID: %v present in docs: %v\n", vecID, docs)
+						for vecID, doc := range vecDocIDMap {
+							fmt.Printf("vector with vecID: %v present in doc: %v\n", vecID, doc)
 						}
 					}
 				case 4:
@@ -116,9 +115,9 @@ var vectorCmd = &cobra.Command{
 	},
 }
 
-func decodeSection(data []byte, start uint64) (int, int, map[int64][]uint32, *faiss.IndexImpl, error) {
+func decodeSection(data []byte, start uint64) (int, int, map[uint64]uint64, *faiss.IndexImpl, error) {
 	pos := int(start)
-	vecDocIDMap := make(map[int64][]uint32)
+	vecDocIDMap := make(map[uint64]uint64)
 
 	// loading doc values - adhering to the sections format. never
 	// valid values for vector section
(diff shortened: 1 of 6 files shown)


section_faiss_vector_index.go
```diff
@@ -619,13 +557,19 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
...
-	var rv int64
+	var rv, sum int64
 	for _, v := range a {
-		rv = int64(math.Float32bits(v)) + (31 * rv)
+		sum += int64(math.Float32bits(v))
 	}
 
+	// Similar to getVectorCode(), this uses the first 32 bits for the vector sum
+	// and the last 32 for a random 32-bit int to ensure identical vectors have
+	// unique hashes.
```
abhinavdangeti, line 571: Not sure i follow your reasoning here, why do the bitwise OR operation below to obtain a unique hash. Is there some significance in retaining the first 32 bits for the vector sum?

section_faiss_vector_index.go
```diff
@@ -619,13 +557,19 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
 
 // todo: better hash function?
 // keep the perf aspects in mind with respect to the hash function.
-// random seed based hash golang.
+// Uses a time based seed to avoid 2 identical vectors in different
+// segments having the same hash -> could be an issue when merging those
+// segments
```
abhinavdangeti, line 562: If you're not going to be de-duplicating vectors from different docIDs - why would there be an issue?
Also - you cannot leave comments like these in here, address them!

section_faiss_vector_index.go
```diff
@@ -657,7 +601,7 @@ type indexContent struct {
 
 type vecInfo struct {
 	vec    []float32
-	docIDs *roaring.Bitmap
+	docIDs uint64
```
abhinavdangeti, line 604: You changed the type but left the name as it were - change to docID and leave it at uint32.

section_faiss_vector_index.go
```diff
@@ -591,11 +531,9 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
 		subVecHash := hashCode(subVec)
 		if _, ok := vo.vecIDMap[subVecHash]; !ok {
 			vo.vecIDMap[subVecHash] = vecInfo{
-				docIDs: roaring.NewBitmap(),
+				docIDs: uint64(docNum),
```
abhinavdangeti, line 534: Why cast a docNum to uint64 here, leave it at uint32. docIDs needs refactoring to docID with your proposal.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
metonymic-smokey: If you're referring to the merge 'issue' in the comment as something to be addressed, it's already addressed when ensuring that no identical vectors have identical hashes. What I meant was that identical vectors having identical hashes was to be avoided during merge.
I think I didn't phrase my comment right.

section_faiss_vector_index.go
metonymic-smokey: On the off chance that the random number generated is the same for 2 different vectors - this adds an additional layer of differentiation.

section_faiss_vector_index.go
metonymic-smokey: Did that to avoid casting here - https://github.com/blevesearch/zapx/blob/511fd66ddea9a1301b99f240d42fadb0c3ee7e16/section_faiss_vector_index.go#L495

faiss_vector_posting.go
```diff
@@ -295,22 +295,32 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (
...
+		for vecID, docID := range vecDocIDMap {
+			if except != nil && except.Contains(docID) {
+				vectorIDsToExclude = append(vectorIDsToExclude, vecID)
+			}
+		}
+
+		var scores []float32
+		var ids []int64
+		var err error
+		scores, ids, err = vecIndex.SearchWithoutIDs(qVector, k, vectorIDsToExclude)
```
Thejas-bhat, line 310: i think `scores, ids, err := vecIndex.SearchWithoutIDs(qVector, k, vectorIDsToExclude)` should suffice over here right?

section_faiss_vector_index.go
```diff
@@ -619,13 +557,19 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
...
-// random seed based hash golang.
+// Uses a time based seed to avoid 2 identical vectors in different
+// segments having the same hash -> could be an issue when merging those
+// segments
 func hashCode(a []float32) int64 {
-	var rv int64
+	var rv, sum int64
 	for _, v := range a {
-		rv = int64(math.Float32bits(v)) + (31 * rv)
+		sum += int64(math.Float32bits(v))
```
Thejas-bhat, line 566: could you please explain why the operation (31 * rv) is dropped? like from what i remember this formula i borrowed from java sdk, and from what i understood this factor was something that helped in generating a better hash than sum of values in an array.

metonymic-smokey: yes, that's the reason - the main differentiator is now the combination of vector sum + random int. 
Even if 2 different vectors sum to the same number, the random int in the other 32 bits should be the differentiator.

cmd/zap/cmd/vector.go
```diff
@@ -139,32 +138,12 @@ func decodeSection(data []byte, start uint64) (int, int, map[int64][]uint32, *fa
 	numVecs, n := binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
 	pos += n
 	for i := 0; i < int(numVecs); i++ {
-		vecID, n := binary.Varint(data[pos : pos+binary.MaxVarintLen64])
+		vecID, n := binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
```
Thejas-bhat, line 141: this should be binary.Varint() right? like i see that in other parts of the code the vecID is retained as binary.Varint()

Thejas-bhat (commented):
just this one comment, and given that you've tested this out, i think the PR makes sense to me.

cmd/zap/cmd/vector.go
metonymic-smokey: yeah, my bad - accidentally changed this for running some edge case tests.

PR #206, opened 2024-01-17
MB-60367: Handle memory allocations more responsibly 

Thejas-bhat:
- when the inverted index section processes the vector fields as well, there are some amount of allocations possible for content which its not supposed to process (for example vector content. This PR fixes that particular short-sight. 
- during the merge process for every field we are currently allocating a couple of slices of capacity = number of segments. these allocations can be reused
- currently the freeing of the reconstructed vector indexes (which can be large in high volume) is a defer logic. This holds up the memory on C heap until the end of function before which there are expensive steps of train and add to the merged index - which will be detrimental overall particularly while introducing the segments as well in high data ingestion scenarios.


section_faiss_vector_index.go
```diff
@@ -34,6 +34,12 @@ import (
 func init() {
 	rand.Seed(time.Now().UTC().UnixNano())
 	registerSegmentSection(SectionFaissVectorIndex, &faissVectorIndexSection{})
+	isInvalidInvertedIndexField = func(field index.Field) bool {
+		if _, ok := field.(index.VectorField); ok {
```
metonymic-smokey, line 38: Nit: could be shortened to :
```
_, ok := field.(index.VectorField)
return ok`

metonymic-smokey:
Looks like there are some conflicts, could you pls resolve them?

abhinavdangeti (approved):

PR #208, opened 2024-01-22
MB-60400: Reduce Garbage generated

CascadingRadium:
All the points below remove garbage to some extent

- Implement caching for MaxDocID in a segment, leveraging the immutability of the FST for the segment. This enhancement is particularly beneficial for the Introducer Loop, where the DocNumbers() API is called for each segment in the root. By storing the MaxDocID and avoiding redundant recomputation using fst.GetMaxKey(). We also cache the ID Dict for a segment to avoid generating the same object multiple times, saving on the space alloced to repeated creations of the FST Reader object.
- Optimize the writeVectorIndexes function by preventing repeated append scenarios. This is achieved by preallocating the necessary memory through an additional for loop, enhancing efficiency.
- Close the enumerator, to free the FST iterators
- Add capacities to many slices to avoid repeatedly appending to the same slice, which generates garbage.

### DIFF
--- a/dict.go
+++ b/dict.go
@@ -16,6 +16,7 @@ package zap
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/RoaringBitmap/roaring"
 	index "github.com/blevesearch/bleve_index_api"
@@ -25,10 +26,15 @@ import (
 
 // Dictionary is the zap representation of the term dictionary
 type Dictionary struct {
-	sb        *SegmentBase
-	field     string
-	fieldID   uint16
-	fst       *vellum.FST
+	sb      *SegmentBase
+	field   string
+	fieldID uint16
+	fst     *vellum.FST
+
+	// if the dictionary is shared across multiple threads
+	// we need to protect the fstReader with a mutex
+	// since it is not thread safe
+	m         sync.Mutex
 	fstReader *vellum.Reader
 
 	bytesRead uint64
@@ -56,7 +62,9 @@ func (d *Dictionary) postingsList(term []byte, except *roaring.Bitmap, rv *Posti
 		return d.postingsListInit(rv, except), nil
 	}
 
+	d.m.Lock()
 	postingsOffset, exists, err := d.fstReader.Get(term)
+	d.m.Unlock()
 	if err != nil {
 		return nil, fmt.Errorf("vellum err: %v", err)
 	}
--- a/merge.go
+++ b/merge.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"os"
 	"sort"
+	"sync"
 
 	"github.com/RoaringBitmap/roaring"
 	seg "github.com/blevesearch/scorch_segment_api/v2"
@@ -31,6 +32,12 @@ import (
 var DefaultFileMergerBufferSize = 1024 * 1024
 
 const docDropped = math.MaxUint64 // sentinel docNum to represent a deleted doc
+
+var mergeBufferPool = sync.Pool{
+	New: func() interface{} {
+		return bufio.NewWriterSize(nil, DefaultFileMergerBufferSize)
+	},
+}
 
 // Merge takes a slice of segments and bit masks describing which
 // documents may be dropped, and creates a new segment containing the
@@ -68,7 +75,8 @@ func mergeSegmentBases(segmentBases []*SegmentBase, drops []*roaring.Bitmap, pat
 	}
 
 	// buffer the output
-	br := bufio.NewWriterSize(f, DefaultFileMergerBufferSize)
+	br := mergeBufferPool.Get().(*bufio.Writer)
+	br.Reset(f)
 
 	// wrap it for counting (tracking offsets)
 	cr := NewCountHashWriterWithStatsReporter(br, s)
@@ -106,7 +114,9 @@ func mergeSegmentBases(segmentBases []*SegmentBase, drops []*roaring.Bitmap, pat
 		return nil, 0, err
 	}
 
-	return newDocNums, uint64(cr.Count()), nil
+	numBytesWritten := cr.Count()
+	mergeBufferPool.Put(br)
+	return newDocNums, uint64(numBytesWritten), nil
 }
 
 func MergeToWriter(segments []*SegmentBase, drops []*roaring.Bitmap,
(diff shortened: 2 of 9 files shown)


section_faiss_vector_index.go
```diff
@@ -438,9 +438,16 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
...
-		var ids []int64
+		// calculate the capacity of the vecs and ids slices
+		// to avoid multiple allocations.
+		vecsLen := 0
+		idsLen := 0
+		for _, vecInfo := range content.vecs {
+			vecsLen += len(vecInfo.vec)
+			idsLen++
+		}
+		vecs := make([]float32, 0, vecsLen)
```
Thejas-bhat, line 449: shouldn't the following suffice?
```
vecs := make([]float32, 0, len(content.vecs) * content.dim)
ids := make([]int64, 0, len(content.vecs))
```

CascadingRadium: bro that vecInfo.vec has the same len always?
like map can be like this?

ABC = [1,2,4]
XYZ = [6,7]

thts y i thought one for loop. If all vecInfo.len is same then i can remiove off this loop

segment.go
```diff
@@ -581,6 +585,28 @@ func (s *SegmentBase) Count() uint64 {
 	return s.numDocs
 }
 
+func (s *SegmentBase) getMaxDocID(idDict *Dictionary) (string, error) {
+	// max key is cached, return it
+	if s.cachedMaxDocID != "" {
```
Thejas-bhat, line 590: shouldn't this be under lock (also is this a duplicate of the following lines)?

section_faiss_vector_index.go
Thejas-bhat: yeah all the vectors' dimensions match at this point, so its not needed

segment.go
abhinavdangeti: Ya, looks like there could be a data race otherwise.

CascadingRadium: sorry

dict.go
```diff
@@ -56,7 +62,9 @@ func (d *Dictionary) postingsList(term []byte, except *roaring.Bitmap, rv *Posti
 		return d.postingsListInit(rv, except), nil
 	}
 
+	d.m.Lock()
```
Thejas-bhat, line 65: can you please explain why this lock is necessary? 
from what i know this path is hit at query time after all the fsts and their readers have been loaded, so at this point all the threads that do access this segment and try to form a postings list would be reading the dictionary - so i dont think a lock is necessary. and there wouldn't be a write - read conflict because the reader is populated after loading the dictionary from the sb.mem[], so this reader's read API won't conflict with the creating and populating the reader inside the dictionary. please let me know if this makes sense

CascadingRadium: Yes that is true
the fstReader is meant for single threaded use, and hence so is the dictionary object
To accomodate this in the query path we essentially create a dictionary object for the field queried from all segments - So for N segments, N dictionary objects and N fst Iterators, because each dictionary create creates a fresh FST Iterator Object.

But the new use case i added is that you cache the dictionary for the "_id" field in the segment - thereby making it a shared object that can be subject to multithreaded access, in which case the lock is needed to allow for single threaded access to the cached "_id" dict's fst reader

This is beacuse from the pprof i could see that for every segment introduced, in a high indexing op scenario,

for an existing segment object S, S's DocNumbers function was called in the introducer, on an avg of around 20 times (after that the segment S just gets deleted due to merge). So the same id dict gets created 20 times, with 20 different FST iterator objects being created as well, which creates a lot of garbage to be GC'd.

CascadingRadium: hmm ok, thing is in the query path the dictionary object is never shared across multiple threads, so the lock is useless in that path.
So i can actually create a new function called postingsListLock() that will only be called by DocNumbers() in the introducer path which is the only place where a doctionary can be shared across threads?

Thejas-bhat: I see yeah now I see that only the idDict that's cached would be shared across threads. 

I think even DocID searcher uses docNumbers() API and if there are parallel deployment of those queries - and while they form the https://github.com/blevesearch/bleve/blob/5f1f45a5c32a140546e247ba91e7250de247adf9/index/scorch/snapshot_index.go#L331 

object parallely there can be shared variable contention over here as well? Won't affect the other types of queries that are there - but in this particular query there can be contention?

CascadingRadium: yes multiple docID queries would use the same dict, so the lock would come into effect there, so then i could seperate this as well, docID queries will call the old API and introducer will call DocNumbersShared, only difference being that from query side (DocIDquery) a new dict is always created and not cached

segment.go
```diff
@@ -581,37 +587,62 @@ func (s *SegmentBase) Count() uint64 {
 	return s.numDocs
 }
 
+func (s *SegmentBase) getDocIDinfo() (*Dictionary, string, error) {
+	// obtain a read lock to check if the max doc ID and dict for _id is cached
+	s.docIDMutex.RLock()
+	cachedDocID := s.cachedMaxDocID
+	cachedDict := s.idDict
+	s.docIDMutex.RUnlock()
```
abhinavdangeti, line 595: This should move to line 596.5 and another unlock after the conditional at 599.5.

dict.go
```diff
@@ -25,10 +26,15 @@ import (
...
-	fst       *vellum.FST
+	sb      *SegmentBase
+	field   string
+	fieldID uint16
+	fst     *vellum.FST
+
+	// if the dictionary is shared across multiple threads
+	// we need to protect the fstReader with a mutex
+	// since it is not thread safe
+	m         sync.Mutex
```
abhinavdangeti, line 37: Sure the reader is not thread safe - why is guarding access here with a mutex a concern now?

CascadingRadium: the dictionary object is always freshly created for any get Dictionary(field) call by a thread. The dictionary object will not be shared across any thread and will remain unique to each thread. With the new approach suggested here, the dictionary for the "_id" field will be created only once and cached in the segment for its lifetime. So it is subject to being shared across threads due to the cache being a shared resource. Hence, here it becomes necessary to guard the fst reader with a mutex.

abhinavdangeti (approved):
Looks fine to me. Merge _only_ after 1 more review at least.

merge.go
```diff
@@ -106,7 +114,9 @@ func mergeSegmentBases(segmentBases []*SegmentBase, drops []*roaring.Bitmap, pat
 		return nil, 0, err
 	}
 
-	return newDocNums, uint64(cr.Count()), nil
+	numBytesWritten := cr.Count()
+	mergeBufferPool.Put(br)
```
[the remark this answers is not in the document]
CascadingRadium: ok i will add section to sheet

abhinavdangeti (approved):

PR #212, opened 2024-02-02
Introducing a size API for vector index on query path

abhinavdangeti:


abhinavdangeti:
It seems unit tests were not addressed.

PR #213, opened 2024-02-05
Early exit condition for vector index merging

metonymic-smokey:
This PR adds an early exit if the vector indexes' merge is aborted/cancelled.
Since merging is an expensive affair, it is worth pre-empting it with such early exits to avoid resource spikes, even if temporary.


section_faiss_vector_index.go
```diff
@@ -259,9 +259,15 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(fieldID int, sbs []*Segme
 
 	vecIndexes := make([]*faiss.IndexImpl, 0, len(sbs))
 	reconsCap := 0
-	for segI, seg := range sbs {
+	for segI, segBase := range sbs {
+		// Considering merge operations on vector indexes are expensive, it is
+		// worth including an early exit if the merge is aborted, saving us
+		// the resource spikes, even if temporary.
+		if isClosed(closeCh) {
+			return seg.ErrClosed
```
Thejas-bhat, line 267: you might need to invoke a freeReconstructedIndexes() over here before the exit

Thejas-bhat (approved):

abhinavdangeti (approved):

PR #216, opened 2024-02-07
Reuse buffers while querying vector indexes

CascadingRadium:
- interface changes to support https://github.com/blevesearch/bleve/pull/1980
- i will close PR if interface change is not allowed

### DIFF
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -266,14 +266,14 @@ func (vpl *VecPostingsIterator) BytesWritten() uint64 {
 
 // vectorIndexWrapper conforms to scorch_segment_api's VectorIndex interface
 type vectorIndexWrapper struct {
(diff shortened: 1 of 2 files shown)


faiss_vector_posting.go
```diff
@@ -266,14 +266,14 @@ func (vpl *VecPostingsIterator) BytesWritten() uint64 {
 
 // vectorIndexWrapper conforms to scorch_segment_api's VectorIndex interface
 type vectorIndexWrapper struct {
-	search func(qVector []float32, k int64, except *roaring.Bitmap) (segment.VecPostingsList, error)
+	search func(qVector []float32, k int64, except *roaring.Bitmap, distanceBuffer []float32, labelBuffer []int64) (segment.VecPostingsList, error)
```
abhinavdangeti, line 269: This is detail bleed across projects.

PR #223, opened 2024-03-11
MB-60914: Zap field specific metrics

Likith101:
- Added an UpdateFieldStats function which will populate the passed map with any field related metrics
 - The map is expected to be already initialized with the appropriate metric names
 - As of now, the only metric added is num_vectors which is obtained by looking at its appropriate field section data
 - This call also is only necessary after a merge because otherwise we can obtain it from the analysis results.
 - With the inclusion of UpdateFieldStats, SegmentBase will now implement the FieldStatsReporter defined in scorch_segment_api

### DIFF
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -404,7 +404,7 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex, 
 	return wrapVecIndex, err
 }
 
-func (sb *SegmentBase) UpdateFieldStats(stats map[string]map[string]int) {
+func (sb *SegmentBase) UpdateFieldStats(stats segment.FieldStats) {
 
 	for _, fieldName := range sb.fieldsInv {
 
@@ -413,15 +413,13 @@ func (sb *SegmentBase) UpdateFieldStats(stats map[string]map[string]int) {
 			continue
 		}
 
-		_, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-		pos += n
-		_, n = binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-		pos += n
-		_, n = binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-		pos += n
+		for i := 0; i < 3; i++ {
+			_, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+			pos += n
+		}
 
 		numVecs, _ := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
 
-		stats["num_vectors"][fieldName] = int(numVecs)
+		stats.AddStat("num_vectors", fieldName, numVecs)
 	}
 }
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -403,3 +403,25 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex, 
 	vecIndex, err = faiss.ReadIndexFromBuffer(indexBytes, faiss.IOFlagReadOnly)
 	return wrapVecIndex, err
 }
+
+func (sb *SegmentBase) UpdateFieldStats(stats map[string]map[string]int) {
+
+	for _, fieldName := range sb.fieldsInv {
+
+		pos := int(sb.fieldsSectionsMap[sb.fieldsMap[fieldName]-1][SectionFaissVectorIndex])
+		if pos == 0 {
+			continue
+		}
+
+		_, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+		pos += n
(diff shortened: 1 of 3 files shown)


faiss_vector_posting.go
```diff
@@ -403,3 +403,25 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex,
 	vecIndex, err = faiss.ReadIndexFromBuffer(indexBytes, faiss.IOFlagReadOnly)
 	return wrapVecIndex, err
 }
+
+func (sb *SegmentBase) UpdateFieldStats(stats map[string]map[string]int) {
```
Thejas-bhat, line 407: i think you can unit test this out

faiss_vector_posting.go
```diff
@@ -403,3 +403,25 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex,
...
+
+	for _, fieldName := range sb.fieldsInv {
+
+		pos := int(sb.fieldsSectionsMap[sb.fieldsMap[fieldName]-1][SectionFaissVectorIndex])
+		if pos == 0 {
+			continue
+		}
+
+		_, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
```
metonymic-smokey, line 416: Think this can be done in a loop.

faiss_vector_posting.go
```diff
@@ -403,3 +403,25 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex,
...
+		pos += n
+		_, n = binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+		pos += n
+		_, n = binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+		pos += n
+
+		numVecs, _ := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+
+		stats["num_vectors"][fieldName] = int(numVecs)
```
metonymic-smokey, line 425: I see you're using "num_vectors" in both the zapx and bleve PRs - I think this can be declared once and re-used.

faiss_vector_posting.go
```diff
@@ -403,3 +403,23 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex,
 	vecIndex, err = faiss.ReadIndexFromBuffer(indexBytes, faiss.IOFlagReadOnly)
 	return wrapVecIndex, err
 }
+
+func (sb *SegmentBase) UpdateFieldStats(stats segment.FieldStats) {
+
+	for _, fieldName := range sb.fieldsInv {
+
```
abhinavdangeti, line 410: Drop this empty line and the one in 408.

faiss_vector_posting.go
```diff
@@ -403,3 +403,23 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex,
...
+		}
+
+		for i := 0; i < 3; i++ {
+			_, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+			pos += n
+		}
+
+		numVecs, _ := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+
+		stats.AddStat("num_vectors", fieldName, numVecs)
```
abhinavdangeti, line 423: AddStat -> Store

See comments on https://github.com/blevesearch/scorch_segment_api/pull/38

faiss_vector_posting.go
Likith101: Added unit tests in bleve which cover this code path as well as merge

faiss_vector_posting.go
Likith101: Added stuff in scorch_segment_api to track the stat names

faiss_vector_posting.go
```diff
@@ -403,3 +403,20 @@ func (sb *SegmentBase) InterpretVectorIndex(field string) (segment.VectorIndex,
 	vecIndex, err = faiss.ReadIndexFromBuffer(indexBytes, faiss.IOFlagReadOnly)
 	return wrapVecIndex, err
 }
+
+func (sb *SegmentBase) UpdateFieldStats(stats segment.FieldStats) {
+	for _, fieldName := range sb.fieldsInv {
+		pos := int(sb.fieldsSectionsMap[sb.fieldsMap[fieldName]-1][SectionFaissVectorIndex])
```
Thejas-bhat, line 409: you can avoid the sb.fieldsMap[fieldName] over here, since I believe the index of fieldsInv is the fieldID. you can verify it and do the change.

Likith101: I'm keeping this the same because there seems to be one case where the fieldsInv looks like this

_id _all content id _id _all content id

So basically theres copies of stuff. I'm not sure where the copies are coming from but for safety reasons (The process will crash with index out of range error otherwise) I'm keeping it in.

abhinavdangeti (approved):

PR #225, opened 2024-03-12
MB-60943 - Add a coarse quantiser to the IVF indexes

metonymic-smokey:
This PR adds a HNSW index as the coarse quantiser for IVF indexes - avoid brute force search for the centroids closest to the query vector and instead use a HNSW centroid index .

Verified this to not impact recall significantly in E2E testing - 
1. 10M dataset - 768 dims
Case | Recall | Accuracy
without change | 0.85795 | 0.897
with change | 0.8565999999999999 | 0.895

2. 5M dataset - 1536 dims
Case | Recall | Accuracy
without change | 0.9591	| 0.979
with change | 0.95811 | 0.981

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -350,8 +350,8 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(sbs []*SegmentBase,
 
 	nvecs := len(finalVecIDs)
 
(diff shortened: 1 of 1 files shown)


abhinavdangeti (approved):
@metonymic-smokey no recall trade-offs I assume with the better latencies? Worth sharing a table comparing the recalls in the description above^.

PR #226, opened 2024-03-12
MB-61029: Deferring the closing of vector index

Thejas-bhat:
- Uses the exponentially weighted moving average to get the average hits on a particular field (thereby a particular vector index). The index stays in the memory when the average of the hits is above a certain threshold and below which the index is closed and the memory is freed for reuse on C side of things.
- This ensures that we are not keeping the index in the memory even when there is no query workload on the field in a segment so the memory pressure does get reduced on the C side of operations.

### DIFF
--- a/build.go
+++ b/build.go
@@ -175,6 +175,9 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32,
 		docValueOffset:      0, // docValueOffsets identified automatically by the section
 		dictLocs:            dictLocs,
 		fieldFSTs:           make(map[uint16]*vellum.FST),
+		vectorCache: &vecCache{
+			cache: make(map[uint16]*cacheEntry),
+		},
 	}
 	sb.updateSize()
 
--- a/build.go
+++ b/build.go
@@ -175,7 +175,7 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32,
 		docValueOffset:      0, // docValueOffsets identified automatically by the section
 		dictLocs:            dictLocs,
 		fieldFSTs:           make(map[uint16]*vellum.FST),
-		vectorCache:         newVectorCache(),
+		vectorCache:         newVectorIndexCache(),
 	}
 	sb.updateSize()
 
--- a/build.go
+++ b/build.go
@@ -175,7 +175,7 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32,
 		docValueOffset:      0, // docValueOffsets identified automatically by the section
 		dictLocs:            dictLocs,
 		fieldFSTs:           make(map[uint16]*vellum.FST),
-		vectorCache:         newVectorIndexCache(),
+		vecIndexCache:       newVectorIndexCache(),
 	}
 	sb.updateSize()
 
--- a/build.go
+++ b/build.go
@@ -175,9 +175,7 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32,
 		docValueOffset:      0, // docValueOffsets identified automatically by the section
 		dictLocs:            dictLocs,
 		fieldFSTs:           make(map[uint16]*vellum.FST),
-		vectorCache: &vecCache{
-			cache: make(map[uint16]*cacheEntry),
-		},
+		vectorCache:         newVectorCache(),
 	}
 	sb.updateSize()
 
--- a/faiss_vector_cache.go
+++ b/faiss_vector_cache.go
@@ -32,15 +32,22 @@ type vecIndexCache struct {
 }
 
 type ewma struct {
-	alpha  float64
-	avg    float64
+	alpha float64
+	avg   float64
+	// every hit to the cache entry is recorded as part of a sample
+	// which will be used to calculate the average in the next cycle of average
+	// computation (which is average traffic for the field till now). this is
+	// used to track the per second hits to the cache entries.
 	sample uint64
 }
 
 type cacheEntry struct {
 	tracker *ewma
-	refs    int64
-	index   *faiss.IndexImpl
+	// this is used to track the live references to the cache entry,
+	// such that while we do a cleanup() and we see that the avg is below a
+	// threshold we close/cleanup only if the live refs to the cache entry is 0.
+	refs  int64
+	index *faiss.IndexImpl
 }
 
 func newVectorIndexCache() *vecIndexCache {
@@ -53,6 +60,11 @@ func newVectorIndexCache() *vecIndexCache {
 func (vc *vecIndexCache) Clear() {
 	vc.m.Lock()
 	close(vc.closeCh)
+
+	// forcing a close on all indexes to avoid memory leaks.
+	for _, entry := range vc.cache {
+		entry.closeIndex()
+	}
 	vc.cache = nil
 	vc.m.Unlock()
 }
@@ -142,7 +154,7 @@ func (vc *vecIndexCache) decRef(fieldIDPlus1 uint16) {
 	vc.m.RUnlock()
 }
 
-func (vc *vecIndexCache) refresh() bool {
+func (vc *vecIndexCache) cleanup() bool {
 	vc.m.Lock()
 	cache := vc.cache
 
@@ -170,14 +182,16 @@ func (vc *vecIndexCache) refresh() bool {
 	return rv
 }
 
+var monitorFreq = 1 * time.Second
+
 func (vc *vecIndexCache) monitor() {
-	ticker := time.NewTicker(1 * time.Second)
+	ticker := time.NewTicker(monitorFreq)
 	for {
 		select {
 		case <-vc.closeCh:
 			return
 		case <-ticker.C:
-			exit := vc.refresh()
+			exit := vc.cleanup()
 			if exit {
 				// no entries to be monitored, exit
 				return
@@ -208,9 +222,6 @@ func initCacheEntry(index *faiss.IndexImpl, alpha float64) *cacheEntry {
 }
 
 func (vc *cacheEntry) incHit() {
-	// every access to the cache entry is accumulated as part of a sample
-	// which will be used to calculate the average in the next cycle of average
-	// computation
 	atomic.AddUint64(&vc.tracker.sample, 1)
 }
 
@@ -223,6 +234,8 @@ func (vc *cacheEntry) decRef() {
 }
 
 func (vc *cacheEntry) closeIndex() {
-	vc.index.Close()
-	vc.index = nil
+	go func() {
+		vc.index.Close()
+		vc.index = nil
+	}()
 }
--- a/faiss_vector_cache.go
+++ b/faiss_vector_cache.go
@@ -26,8 +26,9 @@ import (
 )
 
 type vecCache struct {
-	m     sync.RWMutex
-	cache map[uint16]*cacheEntry
+	closeCh chan struct{}
+	m       sync.RWMutex
+	cache   map[uint16]*cacheEntry
 }
 
 type ewma struct {
@@ -37,8 +38,7 @@ type ewma struct {
 }
 
 type cacheEntry struct {
-	cacheMonitor *ewma
-	closeCh      chan struct{}
+	tracker *ewma
 
 	m     sync.RWMutex
 	index *faiss.IndexImpl
@@ -52,13 +52,10 @@ func newVectorCache() *vecCache {
 
 func (vc *vecCache) Clear() {
 	vc.m.Lock()
-	defer vc.m.Unlock()
+	close(vc.closeCh)
+	vc.cache = nil
+	vc.m.Unlock()
 
-	// close every cache monitor in the cache, thereby closing the index
-	// cached as well.
-	for _, c := range vc.cache {
-		close(c.closeCh)
-	}
 }
 
 func (vc *vecCache) checkCacheForVecIndex(fieldID uint16,
@@ -88,15 +85,6 @@ func (vc *vecCache) isVecIndexCached(fieldID uint16) (*faiss.IndexImpl, bool) {
 	rv := entry.index
 	entry.m.RUnlock()
 
-	// the following cleanup is when the index is closed as part of cacheEntry.closeIndex()
-	// but the corresponding field entry in the vecCache is not cleaned up
-	// fixme: this looks ugly must be refactored in a better way.
-	if rv == nil {
-		vc.m.Lock()
-		delete(vc.cache, fieldID)
-		vc.m.Unlock()
-	}
-
 	return rv, present && (rv != nil)
 }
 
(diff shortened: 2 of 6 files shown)


faiss_vector_cache_noop.go
```diff
@@ -0,0 +1,31 @@
...
+
+package zap
+
+import (
+	"sync"
+)
+
+type vecCache struct {
+	m     sync.RWMutex
+	cache map[uint16]*cacheEntry
```
abhinavdangeti, line 26: Don't really see the need for these attributes here.

segment.go
```diff
@@ -53,8 +53,11 @@ func (*ZapPlugin) Open(path string) (segment.Segment, error) {
 
 	rv := &Segment{
 		SegmentBase: SegmentBase{
-			fieldsMap:      make(map[string]uint16),
-			fieldFSTs:      make(map[uint16]*vellum.FST),
+			fieldsMap: make(map[string]uint16),
+			fieldFSTs: make(map[uint16]*vellum.FST),
+			vectorCache: &vecCache{
+				cache: make(map[uint16]*cacheEntry),
```
abhinavdangeti, line 59: Rather than bleeding the cacheEntry details here, add an API - `newVectorCache(..)` that does this for you where the structs are declared.

faiss_vector_posting.go
```diff
@@ -335,9 +335,8 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
 				return rv, nil
 			},
 			close: func() {
-				if vecIndex != nil {
-					vecIndex.Close()
-				}
+				// skipping the closing for now because the index is cached
+				// todo: subjected to change based on a optimization type flag
```
abhinavdangeti, line 339: Let's not make this an optimization thing - this'll be the new default going forward - applicable to _all_ optimizations.

faiss_vector_cache_noop.go
```diff
@@ -0,0 +1,27 @@
+//  Copyright (c) 2024 Couchbase, Inc.
```
abhinavdangeti, line 1: Let's rename this file to faiss_vector_cache_nosup.go instead.

faiss_vector_cache.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (vc *vecCache) Clear() {
+	vc.m.Lock()
+	close(vc.closeCh)
+	vc.cache = nil
+	vc.m.Unlock()
+
+}
+
+func (vc *vecCache) checkCacheForVecIndex(fieldID uint16,
```
abhinavdangeti, line 62: Rename to simply `loadVectorIndex`.

faiss_vector_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	} else {
+		// if the cache doesn't have vector index, just construct it out of the
+		// index bytes and update the cache.
+		vecIndex, err = faiss.ReadIndexFromBuffer(indexBytes, faiss.IOFlagReadOnly)
+		vc.update(fieldID, vecIndex)
+	}
+	return vecIndex, err
+}
+
+func (vc *vecCache) isVecIndexCached(fieldID uint16) (*faiss.IndexImpl, bool) {
```
abhinavdangeti, line 77: isIndexCached(..)*

faiss_vector_cache.go
```diff
@@ -0,0 +1,206 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+
+	faiss "github.com/blevesearch/go-faiss"
+)
+
+type vecCache struct {
```
abhinavdangeti, line 28: vecIndexCache* for non-ambiguity.

segment.go
```diff
@@ -55,6 +55,7 @@ func (*ZapPlugin) Open(path string) (segment.Segment, error) {
 		SegmentBase: SegmentBase{
 			fieldsMap:      make(map[string]uint16),
 			fieldFSTs:      make(map[uint16]*vellum.FST),
+			vectorCache:    newVectorCache(),
```
abhinavdangeti, line 58: vectorIndexCache: newVectorIndexCache()

build.go
```diff
@@ -175,6 +175,7 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32,
 		docValueOffset:      0, // docValueOffsets identified automatically by the section
 		dictLocs:            dictLocs,
 		fieldFSTs:           make(map[uint16]*vellum.FST),
+		vectorCache:         newVectorCache(),
```
abhinavdangeti, line 178: vectorIndexCache: newVectorIndexCache()

faiss_vector_cache.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+type cacheEntry struct {
+	tracker *ewma
+
+	m     sync.RWMutex
+	index *faiss.IndexImpl
+}
+
+func newVectorCache() *vecCache {
```
abhinavdangeti, line 47: func newVectorIndexCache() *vecIndexCache {

faiss_vector_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	entry.addRef()
+}
+
+func (vc *vecCache) clearField(fieldIDPlus1 uint16) {
+	vc.m.Lock()
+	delete(vc.cache, fieldIDPlus1)
+	vc.m.Unlock()
+}
+
+func (vc *vecCache) refreshEntries() (rv int) {
```
abhinavdangeti, line 106: refresh()*

faiss_vector_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	vc.m.RLock()
+	cache := vc.cache
```
abhinavdangeti, line 108: This is racey isn't it - you're grabbing a reference so a data race could occur in updating this map.

abhinavdangeti (changes requested):

faiss_vector_cache.go
Thejas-bhat: done

faiss_vector_cache.go
```diff
@@ -0,0 +1,195 @@
...
+			atomic.StoreUint64(&entry.tracker.sample, 0)
+			entry.closeIndex()
+			delete(vc.cache, fieldIDPlus1)
+			continue
+		}
+		atomic.StoreUint64(&entry.tracker.sample, 0)
+	}
+
+	rv = len(vc.cache)
+	vc.m.Unlock()
```
abhinavdangeti, line 121: Should redo the perf numbers after increasing this lock context.

faiss_vector_posting.go
```diff
@@ -335,9 +335,8 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
 				return rv, nil
 			},
 			close: func() {
-				if vecIndex != nil {
-					vecIndex.Close()
-				}
+				// skipping the closing because the index is cached and it's being
+				// deferred to a later point of time.
```
abhinavdangeti, line 339: Shouldn't we be decrementing ref here?

faiss_vector_cache.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	_, ok := vc.cache[fieldIDPlus1]
+	if !ok {
+		//  initializing the alpha with 0.4 essentially means that we are favoring
+		// 	the history a little bit more relative to the current sample value.
+		// 	this makes the average to be kept above the threshold value for a
+		// 	longer time and thereby the index to be resident in the cache
+		// 	for longer time.
+		vc.cache[fieldIDPlus1] = initCacheEntry(index, 0.4)
```
abhinavdangeti, line 157: Shouldn't we be setting ref to 1 here?

faiss_vector_cache.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (vc *vecIndexCache) monitor() {
+	ticker := time.NewTicker(1 * time.Second)
+	for {
+		select {
+		case <-vc.closeCh:
+			return
+		case <-ticker.C:
+			numEntries := vc.refresh()
+			if numEntries == 0 {
```
abhinavdangeti, line 133: This can race with update's `if len(vc.cache) == 0 { go vc.monitor() }` below, not sure if this can cause anything deviant though.

faiss_vector_cache.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return rv, present && (rv != nil)
+}
+
+func (vc *vecIndexCache) addRef(fieldIDPlus1 uint16) {
+	vc.m.RLock()
+	entry := vc.cache[fieldIDPlus1]
+	vc.m.RUnlock()
+
+	entry.addRef()
```
abhinavdangeti, line 96: This should be within the lock above, to avoid it racing with line 112 below.

faiss_vector_cache.go
```diff
@@ -0,0 +1,195 @@
...
+	close(vc.closeCh)
+	vc.cache = nil
+	vc.m.Unlock()
+}
+
+func (vc *vecIndexCache) loadVectorIndex(fieldID uint16,
+	indexBytes []byte) (vecIndex *faiss.IndexImpl, err error) {
+	cachedIndex, present := vc.isIndexCached(fieldID)
+	if present {
+		vecIndex = cachedIndex
```
abhinavdangeti, line 65: Isn't there a possibility (in theory) that this vecIndex could be closed by the time the reader gets around to actually using it. This is related to the comment I added below in line 157.

abhinavdangeti (changes requested):
@Thejas-bhat still quite uneasy with the possibility of race-iness within this code.

abhinavdangeti:
@Thejas-bhat I'd appreciate an illustration on how this cache management is being done with ewma tracker - perhaps on the ticket or a document which we can attach to the ticket.

faiss_vector_cache.go
Thejas-bhat: i dont think it can issues over here, because the updating of the vc.cache (which affects the length check over here and the prev mentioned line) happens under a lock.

if the refresh() function acquired the lock then the update would execute only after the vc.cache has been brought to the latest view (which can affect the length of cache if there were old enough entries being cleaned up in refresh()) 

if update acquired the lock first then the refresh() wouldn't execute until it has the latest view of the vc.cache (in which case there could be an extra entry in the vc.cache). 

however, i think instead of returning the numEntries maybe a bool flag as to whether we exit the monitor or not would make it even more robust, please let me know if that makes sense.

faiss_vector_cache.go
Thejas-bhat: yeah likith had been working on something along the same lines - if a query (on one thread) is executing the search i.e its at faissIndex.SearchWithoutIDs() and the monitor thread (theoretically) ends up closing the index because the avg dropped below during that execution we might end up in bad situations. likith's last patch addresses this by tracking the live refs to the index which will be decremented in Close() API.

faiss_vector_cache.go
Thejas-bhat: in the latest patch the ref has been incremented as part of the loadVectorIndex() API (which is parent to this).

faiss_vector_cache.go
Thejas-bhat: so the namings in the latest patch is that 

hit(ewma.sample) -> this is used to track the per second hits to the cache entries - helps in computing the average traffic for the field (which is tracked in ewma.avg field) until now. incHit() is the API which increments this value.

refs -> this is used to track the live references to the cache entry, such that while we do a refresh and we see that the avg is below a threshold we close/cleanup only if the live refs to the cache entry is 0. addRef() and decRef() are the corresponding func calls

abhinavdangeti: That's good info - should be added as commentary where declared in the code.

faiss_vector_cache.go
```diff
@@ -0,0 +1,237 @@
...
+	indexBytes []byte) (*faiss.IndexImpl, error) {
+	vc.m.Lock()
+	defer vc.m.Unlock()
+
+	// when there are multiple threads trying to build the index, guard redundant
+	// index creation by doing a double check and return if already created and
+	// cached.
+	entry, present := vc.cache[fieldID]
+	if present {
+		entry.m.RLock()
```
abhinavdangeti, line 85: I see no real purpose for this RLock.
Also, we should incHit() first before returning entry.index.

faiss_vector_cache.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (vc *vecIndexCache) isIndexCached(fieldID uint16) (*faiss.IndexImpl, bool) {
+	vc.m.RLock()
+	entry, present := vc.cache[fieldID]
+	vc.m.RUnlock()
+	if entry == nil {
+		return nil, false
+	}
+
+	entry.m.RLock()
```
abhinavdangeti, line 124: No need for this lock here.

faiss_vector_cache.go
```diff
@@ -0,0 +1,237 @@
...
+	alpha  float64
+	avg    float64
+	sample uint64
+}
+
+type cacheEntry struct {
+	tracker *ewma
+	refs    int64
+
+	m     sync.RWMutex
```
abhinavdangeti, line 44: Thinking we should lose this mutex, related comments before.

abhinavdangeti (changes requested):

faiss_vector_cache.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (vc *cacheEntry) addRef() {
+	atomic.AddInt64(&vc.refs, 1)
+}
+
+func (vc *cacheEntry) decRef() {
+	atomic.AddInt64(&vc.refs, -1)
+}
+
+func (vc *cacheEntry) closeIndex() {
```
abhinavdangeti, line 225: Rename this to closeIndexLOCKED().

faiss_vector_posting.go
```diff
@@ -335,9 +336,9 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
...
+				sb.vectorCache.decRef(fieldIDPlus1)
 			},
 			size: func() uint64 {
 				if vecIndex != nil {
```
abhinavdangeti, line 344: Initialize a variable vecIndexSize at line 299 above.
Assign vecIndex.Size() to it at the time we load.
So size will simply return vecIndexSize

faiss_vector_cache.go
abhinavdangeti: And it's contents need to be non blocking - because vc.Index.Close() could be long. So -

```
go func() {
    vc.index.Close()
    vc.index = nil
}()
```

faiss_vector_cache.go
```diff
@@ -0,0 +1,228 @@
...
+		atomic.StoreUint64(&entry.tracker.sample, 0)
+	}
+
+	rv := len(vc.cache) == 0
+	vc.m.Unlock()
+	return rv
+}
+
+func (vc *vecIndexCache) monitor() {
+	ticker := time.NewTicker(1 * time.Second)
```
abhinavdangeti, line 174: Add a global var for this.

faiss_vector_cache.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (vc *vecIndexCache) decRef(fieldIDPlus1 uint16) {
+	vc.m.RLock()
+	entry := vc.cache[fieldIDPlus1]
+	entry.decRef()
+	vc.m.RUnlock()
+}
+
+func (vc *vecIndexCache) refresh() bool {
```
abhinavdangeti, line 145: Rename to cleanup perhaps, ties in with the return value here.

faiss_vector_cache.go
```diff
@@ -0,0 +1,228 @@
...
+	return &vecIndexCache{
+		cache:   make(map[uint16]*cacheEntry),
+		closeCh: make(chan struct{}),
+	}
+}
+
+func (vc *vecIndexCache) Clear() {
+	vc.m.Lock()
+	close(vc.closeCh)
+	vc.cache = nil
```
abhinavdangeti, line 56: Think we should force close any open indexes at this point - asynchronously via a go routine.

abhinavdangeti (changes requested):

abhinavdangeti:
Looking almost ready @Thejas-bhat 👍🏼

faiss_vector_cache.go
abhinavdangeti: Maybe the function name doesn't need changing :)

abhinavdangeti (dismissed):

faiss_vector_cache.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (vc *vectorIndexCache) cleanup() bool {
+	vc.m.Lock()
+	cache := vc.cache
+
+	// for every field reconcile the average with the current sample values
+	for fieldIDPlus1, entry := range cache {
+		sample := atomic.LoadUint64(&entry.tracker.sample)
+		entry.tracker.add(sample)
```
abhinavdangeti, line 164: @Thejas-bhat I missed this bit - what are we doing here?

abhinavdangeti: Ah nevermind, I misread this.

abhinavdangeti (approved):

PR #231, opened 2024-04-10
MB-61029: Caching Vec To DocID Map

Likith101:
- Generalized some of the cache functions
 - Cache will include vec to docid mapping as well as one structure to help vec excluded calculation as well as the vec excluded structure
 - Added back the cache mutexes because cache reads will also write back some of the structures depending on the except bitmap

### DIFF
--- a/faiss_vector_cache.go
+++ b/faiss_vector_cache.go
@@ -70,6 +70,19 @@ func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, except *roa
 	return index, vecDocIDMap, vecIDsToExclude, err
 }
 
+func (vc *vectorIndexCache) fetch(fieldID uint16) *cacheEntry {
+	vc.m.RLock()
+	defer vc.m.RUnlock()
+
+	entry, ok := vc.cache[fieldID]
+	if !ok {
+		return nil
+	}
+
+	entry.addRef()
+	return entry
+}
+
 func (vc *vectorIndexCache) createAndCache(fieldID uint16, mem []byte, except *roaring.Bitmap) (
 	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32, vecIDsToExclude []int64, err error) {
 	vc.m.Lock()
@@ -78,8 +91,8 @@ func (vc *vectorIndexCache) createAndCache(fieldID uint16, mem []byte, except *r
 	// when there are multiple threads trying to build the index, guard redundant
 	// index creation by doing a double check and return if already created and
 	// cached.
-	entry, exists := vc.cache[fieldID]
-	if exists {
+	entry, ok := vc.cache[fieldID]
+	if ok {
 		entry.addRef()
 		index, vecDocIDMap = entry.load()
 		if except != nil && !except.IsEmpty() {
@@ -123,11 +136,11 @@ func (vc *vectorIndexCache) createAndCache(fieldID uint16, mem []byte, except *r
 		return nil, nil, nil, err
 	}
 
-	vc.upsertLOCKED(fieldID, index, vecDocIDMap)
+	vc.insertLOCKED(fieldID, index, vecDocIDMap)
 	return index, vecDocIDMap, vecIDsToExclude, nil
 }
 
-func (vc *vectorIndexCache) upsertLOCKED(fieldIDPlus1 uint16,
+func (vc *vectorIndexCache) insertLOCKED(fieldIDPlus1 uint16,
 	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32) {
 	// the first time we've hit the cache, try to spawn a monitoring routine
 	// which will reconcile the moving averages for all the fields being hit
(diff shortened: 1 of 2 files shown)


faiss_vector_cache.go
```diff
@@ -57,21 +64,65 @@ func (vc *vecIndexCache) Clear() {
...
+	vecDocIDMap := vc.vecDocIDMap
+	docIDExcluded := vc.docIDExcluded
+	vecExcluded := vc.vecExcluded
+	vc.m.RUnlock()
+
+	if except != nil {
+		newExcluded := false
+		vc.m.Lock()
+		it := except.Iterator()
+		for it.HasNext() {
```
abhinavdangeti, line 97: I don't think this is a good idea - we're trading compute for resource usage.
The least you should be doing is holding a copy of except and do ..
```
if cacheExcept.Intersects(except) {
``` 
.. to determine if the except bitmap has changed or not.

But even this is probably not worth the trade off of having to determine vecExcluded again in the event except has diverged.

abhinavdangeti (changes requested):
@Likith101 See my in-line comments, in summary I don't think we should be caching anything else but `vecToDocIDMap`.

abhinavdangeti:
@Likith101 Let's also rebase your changes over the base branch, some conflicts have cropped up.

Likith101:
> @Likith101 Let's also rebase your changes over the base branch, some conflicts have cropped up.

I'll rebase once the base branch is merged.

abhinavdangeti:
We've merged the base branch, will review after the rebase then (you can hit edit above to change the base branch to `master`) - make sure only the diff appears in this PR here.

abhinavdangeti:
@Likith101 I feel the base branch you're using here has deviated so much and since you have a commit  that does something and then goes back on it with a following commit - resolving is not quite straight forward.

My recommendation for you here is ..
- start off fresh with your load branch `vecToDocMap` at the tip of master
- make all the changes you need again over it
- **force push your code as a single commit** to this branch on origin

Likith101:
Here are some of the performance improvements that I noticed on my local when trying to test these changes.

| Time Between Queries | Average Latency with No Cache | Only Vector Index Cached | Both Index and Map Cached |
| --- | --- | --- | --- |
| 10ms | 559.47ms | 69.74ms | 29.85ms |
| 25ms | 524.53ms | 66.21ms | 32.15ms |
| 50ms | 525.32ms | 65.93ms | 37.58ms |
| 75ms | 522.78ms | 66.86ms | 40.72ms |
| 100ms | 523.66ms | 68.32ms | 45.74ms |
| 200ms | 524.63ms | 84.30ms | 45.82ms |
| 250ms | 537.42ms | 95.01ms | 47.12ms |
| 300ms | 523.27ms | 97.04ms | 43.64ms |
| 500ms | 523.09ms | 98.45ms | 48.17ms |
| 750ms | 528.64ms | 97.59ms | 46.83ms |
| 1000ms | 549.17ms | 97.93ms | 48.17ms |
| 1500ms | 564.24ms | 286.55ms | 270.62ms |
| 2000ms | 562.52ms | 315.56ms | 289.28ms |
| 2500ms | 564.03ms | 624.43ms | 527.14ms |
| 3000ms | 558.75ms | 556.64ms | 550.42ms |
| 5000ms | 563.99ms | 561.57ms | 566.26ms |

Based on these results, we can see that the cache is cleared out after roughly 2.5 seconds after a single query is hit. A burst of queries will mean that more time is needed to clear the cache, but since the decay is exponential, it will not stay loaded for a very long time. (Roughly 100 queries within 1 second will mean the cache stays for 6-7 seconds). The tests used random queries on sift 1M index.

abhinavdangeti (dismissed):
Looking good to me now.
@Thejas-bhat would you make one pass here as well please.

abhinavdangeti (dismissed):
On second thought - both @Likith101 & @Thejas-bhat should comment on any concerns.

abhinavdangeti (dismissed):

faiss_vector_cache.go
```diff
@@ -18,92 +18,130 @@
...
-	if present {
-		vecIndex = cachedIndex
-		vc.incHit(fieldID)
+func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, except *roaring.Bitmap) (
+	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32, vecIDsToExclude []int64, err error) {
+	entry := vc.fetch(fieldID)
+	if entry != nil {
+		index, vecDocIDMap = entry.load()
+		if except != nil && !except.IsEmpty() {
+			for vecID, docID := range vecDocIDMap {
```
Thejas-bhat, line 61: @Likith101, a small doubt here (for future reference as well). I remember you were caching the vecIDsToExclude as well. I realize that it's removed in the latest patch because it can have compute dent while maintaining it in cases the bitmapts are updated etc. Were you able to gauge how much overhead traversing the map and compute the vecIDsToExclude had? My doubt over here could pointing out something very niche but I suppose it's worth keeping in mind if it does have an overhead.

faiss_vector_cache.go
```diff
@@ -18,92 +18,130 @@
...
 	// cached.
-	entry, present := vc.cache[fieldID]
-	if present {
-		rv := entry.index
-		entry.incHit()
-		return rv, nil
+	entry, ok := vc.cache[fieldID]
+	if ok {
+		entry.addRef()
+		index, vecDocIDMap = entry.load()
```
Thejas-bhat, line 97: maybe we can keep the following snippet in a func since its repeated at a couple of places (line 60) calling it "getInvalidVecs" or something better.

faiss_vector_cache.go
```diff
@@ -18,92 +18,130 @@
...
+	defer vc.m.RUnlock()
+
+	entry, ok := vc.cache[fieldID]
+	if !ok {
+		return nil
 	}
-	vc.addRef(fieldID)
-	return vecIndex, err
+
+	entry.addRef()
```
Thejas-bhat, line 82: I see that an entry.addRef() is followed immediately entry.load(). for better readability maybe we can move the addRef() invocation inside the load() function and although it might mean have a RLock() there i dont think it would cause functional defects. Please think upon the same and see if you think it's worth doing it

faiss_vector_cache.go
Likith101: The overhead would only show up when there are constant mutations in the index altering the except bitmap. I was not able to determine how much difference both the approaches made using my testing.

faiss_vector_cache.go
Likith101: It might be better to have the addRef with load. I don't think the locks affect anything here, but I'll check once before making this change

abhinavdangeti (dismissed):
Thanks @Likith101 . Looks good.
Best let @Thejas-bhat put in his review as well to make sure all his comments/concerns have been addressed.

abhinavdangeti (approved):

Thejas-bhat (approved):

PR #233, opened 2024-04-12
minor optimizations and bug fixes

CascadingRadium:
- Ensure slice capacities are accurately calculated to always be equal to the final slice length
- Fix test to cleanup zap segment after completion


abhinavdangeti:
@CascadingRadium would you add to the description above everything this change intends to do.

abhinavdangeti (approved):

section_faiss_vector_index.go
```diff
@@ -275,6 +275,8 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(sbs []*SegmentBase,
 	indexes []*vecIndexMeta, w *CountHashWriter, closeCh chan struct{}) error {
 
 	vecIndexes := make([]*faiss.IndexImpl, 0, len(sbs))
+	finalVecIDCap := 0
```
Thejas-bhat, line 278: nit: you can initialise both these vars in a single line as `var finalVecIDCap, indexDataCap int`

PR #234, opened 2024-04-19
MB-60943 - Reduce number of centroids for IVF indexes.

metonymic-smokey:
This PR reduces the number of centroids for vectors in sizes ranging from 200k to 1M. Since the reduction in training time is linearly dependent on reduction in centroids, this is to drive down training time.


abhinavdangeti (commented):
@metonymic-smokey is this ready for review - would you update the title here to better reflect the change itself.

abhinavdangeti (approved):

PR #239, opened 2024-05-02
bug fix: de-duplicating fieldsInv value

Thejas-bhat:
- there was a logical flaw introduced that resulted in `fieldsInv` having duplicate entries for each field, when it should be a set of unique values.


build.go
```diff
@@ -166,7 +166,6 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32,
 		memCRC:              memCRC,
 		chunkMode:           chunkMode,
 		fieldsMap:           fieldsMap,
-		fieldsInv:           fieldsInv,
```
abhinavdangeti, line 169: It seems we re-init `fieldsMap` as well when we loadFieldsNew(), but at least that would be an overwrite as opposed to an append like with `fieldsInv` - this would not cause any functional issues though would it?

Thejas-bhat: yeah fieldsMap doesn't have a functional issue. i've checked in my custom logs and it turns out that the field:fieldID mapping is retained throughout and there is no change to it. its only the fieldsInv that had the defect.

abhinavdangeti (approved):

PR #246, opened 2024-06-05
MB-62167: Fix windows crash

CascadingRadium:
- Faiss does not support OnDiskInvertedLists on windows, and hence while using go-faiss's IO flags, we must take into account the build platform.
- This PR enforces the old faiss.IOFlagReadOnly flag for windows and maintains the existing io flags for unix based platforms.

### DIFF
--- a/faiss_vector_io_flags_unix.go
+++ b/faiss_vector_io_flags_unix.go
@@ -1,0 +1,22 @@
+//  Copyright (c) 2024 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


faiss_vector_io_flags_unix.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+//go:build vectors && !windows
+// +build vectors,!windows
+
+package zap
+
+import faiss "github.com/blevesearch/go-faiss"
+
+const vectorIndexIOFlags = faiss.IOFlagReadMmap | faiss.IOFlagSkipPrefetch
```
Thejas-bhat, line 22: maybe `faissIOFlags` sounds concise and better?

CascadingRadium: thanks for the review bro 🙏🏻 🙏🏻

abhinavdangeti (approved):

PR #251, opened 2024-07-09
MB-62354: Support Cosine Similarity

CascadingRadium:
- Set FAISS vector metric to be `MetricInnerProduct`, if the distance metric from bleve is either `Cosine` or `InnerProduct`

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -485,18 +485,27 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
 	//    2. its constituent vectorID -> {docID} mapping.
 	tempBuf := vo.grabBuf(binary.MaxVarintLen64)
 	for fieldID, content := range vo.vecFieldMap {
+		// Set the faiss metric type (default is Euclidean Distance or l2_norm)
+		var metric = faiss.MetricL2
+		// Flag to indicate if the vectors need to be normalized before indexing.
+		var normalize bool
+		if content.metric == index.InnerProduct || content.metric == index.CosineSimilarity {
+			// use the same FAISS metric for inner product and cosine similarity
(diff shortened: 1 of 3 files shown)


section_faiss_vector_index.go
```diff
@@ -485,20 +485,29 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
...
+			normalize = content.metric == index.CosineSimilarity
+		}
 		// calculate the capacity of the vecs and ids slices
 		// to avoid multiple allocations.
 		vecs := make([]float32, 0, len(content.vecs)*int(content.dim))
 		ids := make([]int64, 0, len(content.vecs))
 		for hash, vecInfo := range content.vecs {
+			if normalize {
+				// normalize the vector
+				vecInfo.vec = index.NormalizeVector(vecInfo.vec)
```
abhinavdangeti, line 505: Why are we doing this in zap? This should be re-situated within bleve.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
CascadingRadium: i had initially added it here so that is would be agnostic to the type of vector in bleve -> vector / vector base64. When it reaches zap both will be a normal vector only ...

CascadingRadium: ill add it in bleve

abhinavdangeti: Yes please, because vector_base64 decoding is a "bleve" concept as well.

abhinavdangeti (dismissed):

abhinavdangeti:
Do not merge until we've included the go.mod change required here as well. Pending new tag.

abhinavdangeti (approved):

PR #252, opened 2024-07-24
MB-62427: seeking the pointer correctly while reading segmentBase data

Thejas-bhat:
v16 file format introduced the concept of sections in which as part of metadata the `fieldsIndex` part was replaced by a `sectionsIndex` which serves the same purpose of `fieldsIndex` of pointing to a file offset where the actual data of field's section is stored. 

The PR tries to resolve the situation - where there was only one field's data being written out while creating the in-memory `segmentBase` and the process was crashing while initializing the fields related information in the loadFieldsNew() API at https://github.com/blevesearch/zapx/blob/0c7027f136973ae50dd8c86026c557e0a9cfbcf3/segment.go#L323

The reason for the crash is while writing out the data, because there is only one field metadata being written out at the very end of the buffer https://github.com/blevesearch/zapx/blob/0c7027f136973ae50dd8c86026c557e0a9cfbcf3/write.go#L89 is worth 9 bytes. However we were doing an out-of-bounds access over here https://github.com/blevesearch/zapx/blob/0c7027f136973ae50dd8c86026c557e0a9cfbcf3/segment.go#L323 because binary.MaxVarintLen64 = 10.

The solution proposed here is to guard this kind of access by keeping in mind the length of the backing buffer like how it was done in v15 https://github.com/blevesearch/zapx/blob/a72794cd37c5c992f48571b9c6267759cbc5ac33/segment.go#L268

### DIFF
--- a/segment.go
+++ b/segment.go
@@ -322,8 +322,6 @@ func (s *SegmentBase) loadFieldsNew() error {
 		return s.loadFields()
 	}
 
-	fmt.Println("pos:", pos, "pos+binary.MaxVarintLen64:", pos+binary.MaxVarintLen64, "capacity of buffer:", cap(s.mem))
-
 	seek := pos + binary.MaxVarintLen64
 	if seek > uint64(len(s.mem)) {
 		// handling a buffer overflow case.
--- a/segment.go
+++ b/segment.go
@@ -322,8 +322,20 @@ func (s *SegmentBase) loadFieldsNew() error {
 		return s.loadFields()
 	}
 
+	fmt.Println("pos:", pos, "pos+binary.MaxVarintLen64:", pos+binary.MaxVarintLen64, "capacity of buffer:", cap(s.mem))
+
+	seek := pos + binary.MaxVarintLen64
+	if seek > uint64(len(s.mem)) {
+		// handling a buffer overflow case.
+		// a rare case where the backing buffer is not large enough to be read directly via
+		// a pos+binary.MaxVarinLen64 seek. For eg, this can happen when there is only
+		// one field to be indexed in the entire batch of data and while writing out
+		// these fields metadata, you write 1 + 8 bytes whereas the MaxVarintLen64 = 10.
+		seek = uint64(len(s.mem))
+	}
+
 	// read the number of fields
-	numFields, sz := binary.Uvarint(s.mem[pos : pos+binary.MaxVarintLen64])
+	numFields, sz := binary.Uvarint(s.mem[pos:seek])
 	pos += uint64(sz)
 	s.incrementBytesRead(uint64(sz))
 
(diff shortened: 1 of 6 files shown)


abhinavdangeti (approved):

segment.go
```diff
@@ -322,8 +322,18 @@ func (s *SegmentBase) loadFieldsNew() error {
...
+		// one field to be indexed in the entire batch of data and while writing out
+		// these fields metadata, you write 1 + 8 bytes whereas the MaxVarintLen64 = 10.
+		seek = uint64(len(s.mem))
+	}
+
 	// read the number of fields
-	numFields, sz := binary.Uvarint(s.mem[pos : pos+binary.MaxVarintLen64])
+	numFields, sz := binary.Uvarint(s.mem[pos:seek])
 	pos += uint64(sz)
 	s.incrementBytesRead(uint64(sz))
```
abhinavdangeti, line 339: @Thejas-bhat there's no chance of an overflow downstream now is there - where we do this below after updating `pos` ..
```
addr := binary.BigEndian.Uint64(s.mem[pos : pos+8])
```

abhinavdangeti: If you have a reason that we won't hit another overflow downstream, it's worth adding that comment in line with the code.

Thejas-bhat: @abhinavdangeti, here's why i think there won't be panics down the line - in this particular situation, when the buffer since we've established that there are 1 + 8 bytes after the sectionsIndexOffset doing a uvarint read from `pos : len(s.mem) ` leads to only 1 byte being read the first time (from what i've seen in the custom loggings and some experimentation with binary.Uvarint() go.dev playground) this uncovers value of the numFields to be equal to 1 after which the following loop will be performed once to read the field's offset value

Thejas-bhat: at this point, we've read till the eof point of the buffer and the field offset precedes this metadata of field offsets we store. so it starts reading the section from a point where i think the out-of-bounds conditions won't occur (since the existing logic of parsing the inverted text index was used which is stable)

abhinavdangeti: If numFields is equal to one, then like you say - the loop will run that one time. Are you implying that now in line 343 (right side) a new section/buffer is being interpreted?

Would you extend the commentary to what you've added above explaining why no other out-of-bounds could occur in the immediate aftermath.

Thejas-bhat: yeah, true the loop will get executed and hit 343. however the pos would have incremented to sectionsIndexOffset + 1 (and not sectionsIndexOffset + len(s.mem) + 1) because the binary.Uvarint() returns sz = 1 and the line below it does a pos = pos + 1. yeah i will add a commentary over there 👍

abhinavdangeti (approved):

PR #255, opened 2024-08-12
MB-62230 - Support for pre-filtering with kNN

metonymic-smokey:
1. Accommodates filtered doc IDs, if required, within a kNN search over a vector index.
2. Builds and caches a document to vector ID map for looking up vector IDs of the filtered doc IDs.


faiss_vector_posting.go
```diff
@@ -318,12 +325,41 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
...
+						vectorIDsToExclude, params)
+					if err != nil {
+						return nil, err
+					}
+				} else if eligibleDocIDs.Stats().Cardinality > 0 {
+					// Non-zero documents eligible per the filter query.
+
+					// Making doubly sure that the deleted docs do not make their
+					// way into the "inclusion" list, even though that's been
+					// accounted for when computing the eligible doc IDs.
```
abhinavdangeti, line 343: Why the double work? We're controlling this list - re-computation will cost us.

faiss_vector_posting.go
```diff
@@ -318,12 +325,41 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
...
+					eligibleDocIDs.AndNot(except)
+
+					// vector IDs corresponding to the local doc numbers to be
+					// considered for the search
+					vectorIDsToInclude := make([]int64, eligibleDocIDs.Stats().Cardinality)
```
abhinavdangeti, line 348: Lot of allocations here - which is not good ..
* `make` of an array of int64s
* `ToArray()` below allocates an array as well

Ideally just ToArray() should've sufficed, but if you absolutely cannot work with the return type, then you should just iterate over the bitmap to populate `vectorIDsToInclude` as opposed to converting it and then reconverting it - which is quite bad!

abhinavdangeti (changes requested):

faiss_vector_posting.go
```diff
@@ -318,12 +325,41 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
...
 				}
 
-				scores, ids, err := vecIndex.SearchWithoutIDs(qVector, k, vectorIDsToExclude, params)
-				if err != nil {
-					return nil, err
+				var scores []float32
+				var ids []int64
+				var err error
+
+				if !requireFiltering {
```
abhinavdangeti, line 332: Lose this setting and replace with `len(eligibielDocIDs) == 0`

faiss_vector_posting.go
metonymic-smokey: VectorIDs are int64 hashes of the original float32 vector. Hence, the conversion.

faiss_vector_posting.go
```diff
@@ -288,20 +298,40 @@ func (i *vectorIndexWrapper) Size() uint64 {
 // InterpretVectorIndex returns a construct of closures (vectorIndexWrapper)
 // that will allow the caller to -
 // (1) search within an attached vector index
-// (2) close attached vector index
-// (3) get the size of the attached vector index
-func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap) (
+// (2) search limited to a subset of documents within an attached vector index
+// (3) close attached vector index
+// (4) get the size of the attached vector index
+func (sb *SegmentBase) InterpretVectorIndex(field string, filtered bool,
```
abhinavdangeti, line 304: If we're going to need this variable, let's keep the naming consistent (even across repositories) - `filtered` or `requiresFiltering` (whichever makes most sense).

faiss_vector_cache.go
```diff
@@ -52,34 +52,56 @@ func (vc *vectorIndexCache) Clear() {
...
-func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, except *roaring.Bitmap) (
-	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32, vecIDsToExclude []int64, err error) {
+func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte,
+	loadDocVecIDMap bool, except *roaring.Bitmap) (
+	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32, docVecIDMap map[uint32]int64,
+	vecIDsToExclude []int64, err error) {
 	var found bool
-	index, vecDocIDMap, vecIDsToExclude, found = vc.loadFromCache(fieldID, except)
+	index, vecDocIDMap, docVecIDMap, vecIDsToExclude, found = vc.loadFromCache(
+		fieldID, except)
```
abhinavdangeti, line 61: Hmm, something doesn't feel right here - I think it is the situation if the first loadFromCache didn't need us to "loadDocVecIDMap" and while this is still in cache the 2nd invocation requires it .. in which case shouldn't you need to invalidate the cache and rebuild it?

faiss_vector_cache.go
```diff
@@ -52,34 +52,56 @@ func (vc *vectorIndexCache) Clear() {
 	vc.m.Unlock()
 }
 
-func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, except *roaring.Bitmap) (
-	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32, vecIDsToExclude []int64, err error) {
+func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte,
+	loadDocVecIDMap bool, except *roaring.Bitmap) (
```
abhinavdangeti, line 56: Let's add some commentary for this method around when `loadDocVecIDMap` will be expected and when it won't.

faiss_vector_posting.go
abhinavdangeti: Rename to `requiresFiltering` per your `blevesearch/scorch_segment_api` PR.

faiss_vector_posting.go
```diff
@@ -318,23 +348,61 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
...
+
+					// vector IDs corresponding to the local doc numbers to be
+					// considered for the search
+					vectorIDsToInclude := make([]int64, len(eligibleDocIDs))
+					for i, id := range eligibleDocIDs {
+						vectorIDsToInclude[int64(i)] = docVecIDMap[uint32(id)]
+					}
+
+					scores, ids, err = vecIndex.SearchWithIDs(qVector, k,
+						vectorIDsToInclude, params)
```
abhinavdangeti, line 399: Shouldn't you be removing `vectorIDsToExclude` from `vectorIDsToInclude` before this or are we certain there never really is going to be an overlap there because of the pre-filter?

abhinavdangeti (changes requested):

faiss_vector_posting.go
metonymic-smokey: As per my understanding, deleted results aren't returned as part of the initial filter search, whose results form the basis of the vector include list.

faiss_vector_cache.go
metonymic-smokey: good catch, thanks!

abhinavdangeti: Thinking it make be a good idea to integrate loadFromCache and createAndCache into one method - loadFromCache which will also be responsible for creation if it's a cache miss. This would avoid us having to check `!found || (loadDocVecIDMap && docVecIDMap == nil)` twice otherwise.

Do you see any obstacles with this approach?

faiss_vector_cache.go
```diff
@@ -52,34 +52,62 @@ func (vc *vectorIndexCache) Clear() {
...
+	index, vecDocIDMap, docVecIDMap = entry.load()
 	vecIDsToExclude = getVecIDsToExclude(vecDocIDMap, except)
 
-	return index, vecDocIDMap, vecIDsToExclude, true
+	return index, vecDocIDMap, docVecIDMap, vecIDsToExclude, true
 }
 
-func (vc *vectorIndexCache) createAndCache(fieldID uint16, mem []byte, except *roaring.Bitmap) (
-	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32, vecIDsToExclude []int64, err error) {
+func (vc *vectorIndexCache) loadDocVecIDFromCache(fieldID uint16) (
```
abhinavdangeti, line 92: Don't see this method being used anywhere?

faiss_vector_cache.go
metonymic-smokey: none as yet, sounds good.

metonymic-smokey: I think this approach still has the same downside as the earlier approach - the first pre-filtering query will require rebuilding everything other than the docVecIDMap.
I'll try to refactor so that the first pre-filtered query only constructs the docVecIDMap using the vecDocIDMap.

faiss_vector_posting.go
```diff
@@ -318,23 +348,61 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
 					return rv, nil
 				}
 
-				scores, ids, err := vecIndex.SearchWithoutIDs(qVector, k, vectorIDsToExclude, params)
+				var scores []float32
+				var ids []int64
```
Thejas-bhat, line 352: i think these 3 three vars can be init'd as part of 355 itself, since there isn't an if condition around SearchWithoutIDs()

faiss_vector_cache.go
```diff
@@ -52,34 +52,83 @@ func (vc *vectorIndexCache) Clear() {
...
 
 	entry, ok := vc.cache[fieldID]
-	if !ok {
-		return nil, nil, nil, false
+	// either do not require the docVecIDMap or if we do, it's already loaded in
+	// the cache.
+	if ok && (!loadDocVecIDMap || (loadDocVecIDMap && entry.docVecIDMap != nil)) {
+		index, vecDocIDMap, docVecIDMap = entry.load()
+		vecIDsToExclude = getVecIDsToExclude(vecDocIDMap, except)
+		return index, vecDocIDMap, docVecIDMap, vecIDsToExclude, nil
```
abhinavdangeti, line 83: RUnlock's context needs to end here, and a write lock is needed for everything that follows. You can refactor the existing createAndCache to createAndCacheLOCKED to achieve this and it seems you must check for loadDocVecIDMap within right lock again unfortunately.

abhinavdangeti (changes requested):

faiss_vector_cache.go
metonymic-smokey: refactored this part, pls take a look at recent commits.

faiss_vector_cache.go
metonymic-smokey: my bad, missed it the first time.
I've changed that in my recent commits - modifying the cache is done only after acquiring a lock, reading without modifying done under a read lock.

faiss_vector_cache.go
```diff
@@ -52,54 +52,84 @@ func (vc *vectorIndexCache) Clear() {
...
-	if !ok {
-		return nil, nil, nil, false
-	}
-
-	index, vecDocIDMap = entry.load()
-	vecIDsToExclude = getVecIDsToExclude(vecDocIDMap, except)
+	if ok {
+		vecIDsToExclude = getVecIDsToExclude(vecDocIDMap, except)
+		index, vecDocIDMap, docVecIDMap = entry.load()
+		if !loadDocVecIDMap || (loadDocVecIDMap && len(entry.docVecIDMap) == 0) {
```
abhinavdangeti, line 79: Shouldn't this be ..
```
loadDocVecIDMap && len(entry.docVecIDMap) > 0
```

faiss_vector_cache.go
```diff
@@ -52,54 +52,84 @@ func (vc *vectorIndexCache) Clear() {
...
-	// when there are multiple threads trying to build the index, guard redundant
-	// index creation by doing a double check and return if already created and
-	// cached.
-	entry, ok := vc.cache[fieldID]
-	if ok {
-		index, vecDocIDMap = entry.load()
-		vecIDsToExclude = getVecIDsToExclude(vecDocIDMap, except)
-		return index, vecDocIDMap, vecIDsToExclude, nil
+func (vc *vectorIndexCache) addDocVecIDMapToCacheLOCKED(ce *cacheEntry) map[uint32]int64 {
+
```
abhinavdangeti, line 102: Looks this empty line.

faiss_vector_cache.go
```diff
@@ -52,54 +52,84 @@ func (vc *vectorIndexCache) Clear() {
...
+	// Guarding against concurrent accesses by adding a double check here.
```
abhinavdangeti, line 103: Rephrase this to ..
```
// Handle concurrent accesses (to avoid unnecessary work) by adding a check within the write lock here.
```

abhinavdangeti (commented):
Couple minor comments, but looking good to me @metonymic-smokey .

abhinavdangeti (dismissed):

abhinavdangeti:
@metonymic-smokey no more force pushing here please :)

abhinavdangeti (approved):

PR #256, opened 2024-08-29
Introducing config at zap layer

Thejas-bhat:


segment.go
```diff
@@ -38,8 +38,18 @@ func init() {
 	reflectStaticSizeSegmentBase = int(unsafe.Sizeof(sb))
 }
 
+// Open returns a zap impl of a segment along with keeping some config values in
+// mind during the segment's lifetime.
```
metonymic-smokey, line 42: Nit: I think the "in mind" can be removed since we're dealing with segments here :)

abhinavdangeti (approved):
Go ahead and propose this commit for the other 6 versions of zapx also would you.

PR #258, opened 2024-09-03
MB-59633: Remove encoded polygons from term dictionaries

Likith101:
Targeted for v2.5.0


section_inverted_text_index.go
```diff
@@ -558,9 +559,11 @@ func (io *invertedIndexOpaque) writeDicts(w *CountHashWriter) (dictOffsets []uin
 			}
 
 			if postingsOffset > uint64(0) {
-				err = io.builder.Insert([]byte(term), postingsOffset)
-				if err != nil {
-					return nil, err
+				if !strings.HasPrefix(term, "##") {
```
abhinavdangeti, line 562: Add some commentary around here on what the `##` prefix is for, will you.

section_inverted_text_index.go
```diff
@@ -558,9 +559,12 @@ func (io *invertedIndexOpaque) writeDicts(w *CountHashWriter) (dictOffsets []uin
 			}
 
 			if postingsOffset > uint64(0) {
-				err = io.builder.Insert([]byte(term), postingsOffset)
-				if err != nil {
-					return nil, err
+				// Ignore terms with ##. They are glue bytes for encoded polygons
+				if !strings.HasPrefix(term, "##") {
```
CascadingRadium, line 563: this will falsely flag any term with the `##` prefix. So if i index a term with `simple` analyzer and the term is `##zapx` it will not index the term.

Likith101: True, perhaps we could add a length limit here as well as a suffix check. So that only terms with `##asddsa........(100 chars)##` will be filtered. There could still be false positives even with this approach though.

Thejas-bhat: you could perhaps refactor some upstream code to tracking some more info mapping to a fieldID (such as an enum which indicates the field type) and fetch the field type here (=geotype) to make your decision here.

CascadingRadium (dismissed):

section_inverted_text_index.go
```diff
@@ -967,6 +979,8 @@ type invertedIndexOpaque struct {
 
 	tmp0 []byte
 
+	geoShapeFields map[uint16]struct{}
```
Thejas-bhat, line 982: rather than this, I think its better to extend the `fieldsInv` to a struct say fieldInfo or something which will have name and type within it. 

in the getOrDefineField() API you could populate the fieldsInv entry with the type value as well.

That way it would be more future proof (other field types can also leverage this in future if a need comes out) and we don't have the need for an extra member variable either (so lesser objects to maintain by realloc'ing and stuff). 

Please let me know if this sounds better to you all, thanks.

Likith101: I switched the approach so that we store these terms separately and add them only to doc values, let me know if this is better.

section_inverted_text_index.go
```diff
@@ -958,6 +978,9 @@ type invertedIndexOpaque struct {
 	numTermsPerPostingsList []int // key is postings list id
 	numLocsPerPostingsList  []int // key is postings list id
 
+	// docNum -> fieldID -> term
+	specialTerms map[int]map[int][]byte
```
CascadingRadium, line 982: since this has only one term that must be added to docValues, rename to `extraDocValueTerm`

section_inverted_text_index.go
```diff
@@ -789,6 +794,13 @@ func (i *invertedIndexOpaque) realloc() {
 		if field.Options().IncludeDocValues() {
 			i.IncludeDocValues[fieldID] = true
 		}
+
+		if f, ok := field.(index.GeoShapeField); ok {
```
Thejas-bhat, line 798: please correct me if i'm wrong here, but this condition block is hit after getting the `analyzedTokenFrequencies` and tracking as part of the postings. So wouldn't the latest patch write the encoded shape to both the term dictionary (since i dont see any checks avoid an insert to the term dictionary downstream) and the doc values section?

Likith101: There is a bleve change related to this MB (https://github.com/blevesearch/bleve/pull/2162) that removes the term from the dictionaries as well as the token frequencies.

Thejas-bhat (approved):

CascadingRadium (approved):

section_inverted_text_index.go
```diff
@@ -797,14 +809,20 @@ func (i *invertedIndexOpaque) realloc() {
 		i.IncludeDocValues = make([]bool, len(i.FieldsInv))
 	}
 
-	for _, result := range i.results {
+	if i.extraDocValues == nil {
+		i.extraDocValues = map[int]map[int][]byte{}
```
abhinavdangeti, line 813: We're creating garbage here that is applicable to geo shape fields _only_, so 2 points here ..

1. the name feels somewhat generic, and should perhaps state what it applies to - like `geoDocValues`?
2. we should move this to line 798.5 above to limit this allocation to geoshape fields only.

abhinavdangeti (changes requested):
@Likith101 do we have confirmation from QE / perf on things looking ok with this change?

section_inverted_text_index.go
Likith101: I had kept the name generic so that we can reuse this field for other field types if necessary in the future.

abhinavdangeti: It's an un-exported attribute, but ok - would you add some commentary around where the var is declared (line 980 below) indicating what it's currently used for.

abhinavdangeti:
@Likith101 if you could add some more commentary for the map per our conversation^ and testing shows this ok, I'm good to merge this change 👍🏼

abhinavdangeti (approved):

CascadingRadium (approved):

section_inverted_text_index.go
```diff
@@ -958,6 +976,11 @@ type invertedIndexOpaque struct {
 	numTermsPerPostingsList []int // key is postings list id
 	numLocsPerPostingsList  []int // key is postings list id
 
+	// store terms that are unnecessary for the term dictionaries but needed in doc values
+	// eg - encoded geoshapes
+	// docNum -> fieldID -> term
+	extraDocValues map[int]map[int][]byte
```
abhinavdangeti, line 982: Let's change the type of this map to ..
```
extraDocValues map[int]map[uint16][]byte
```

.. to avoid having to re-typecast `fieldID` back to an int in line 802 above.

abhinavdangeti (changes requested):

abhinavdangeti (approved):

CascadingRadium (approved):

Issue #209, opened 2024-01-24
runtime error: integer divide by zero


streamer45:
We received a stack trace for a crash on this library:

```
runtime.errorString: runtime error: integer divide by zero
  File "github.com/blevesearch/zapx/v15@v15.3.11/posting.go", line 747, in (*PostingsIterator).nextDocNumAtOrAfterClean
  File "github.com/blevesearch/zapx/v15@v15.3.11/posting.go", line 632, in (*PostingsIterator).nextDocNumAtOrAfter
  File "github.com/blevesearch/zapx/v15@v15.3.11/posting.go", line 537, in (*PostingsIterator).nextAtOrAfter
  File "github.com/blevesearch/zapx/v15@v15.3.11/posting.go", line 526, in (*PostingsIterator).Next
  File "github.com/blevesearch/bleve/v2@v2.3.8/index/scorch/snapshot_index_tfr.go", line 88, in (*IndexSnapshotTermFieldReader).Next
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_term.go", line 89, in (*TermSearcher).Next
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_conjunction.go", line 218, in (*ConjunctionSearcher).Next
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_conjunction.go", line 246, in (*ConjunctionSearcher).Advance
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_disjunction_slice.go", line 241, in (*DisjunctionSliceSearcher).Advance
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_conjunction.go", line 253, in (*ConjunctionSearcher).advanceChild
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_conjunction.go", line 241, in (*ConjunctionSearcher).Advance
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_boolean.go", line 356, in (*BooleanSearcher).Advance
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_conjunction.go", line 253, in (*ConjunctionSearcher).advanceChild
  File "github.com/blevesearch/bleve/v2@v2.3.8/search/searcher/search_conjunction.go", line 200, in (*ConjunctionSearcher).Next
[...]

PR #263, opened 2024-09-12
Handle zero chunk size

moshaad7:
The existing implementation of the method
getChunkSize(...) (uint64, error), in some cases,
can return chunkSize==0 and err==nil.

The onus is on the caller to check for such possibility and handle it properly.
Callers often use the returned chunkSize as a divisor and a zero chunkSize lead to panic.
see https://github.com/blevesearch/zapx/issues/209

This PR intends to update the method implementation to always return an error in case the returned chunkSize value is 0.
That way caller need to only worry about error being non-nil.

Callers which are ok with 0 chunkSize can check the returned error against ErrChunkSizeZero

### DIFF
--- a/posting.go
+++ b/posting.go
@@ -305,10 +305,7 @@ func (rv *PostingsList) read(postingsOffset uint64, d *Dictionary) error {
 	chunkSize, err := getChunkSize(d.sb.chunkMode,
 		rv.postings.GetCardinality(), d.sb.numDocs)
 	if err != nil {
-		return err
-	} else if chunkSize == 0 {
-		return fmt.Errorf("chunk size is zero, chunkMode: %v, numDocs: %v",
-			d.sb.chunkMode, d.sb.numDocs)
+		return fmt.Errorf("failed to get chunk size: %v", err)
 	}
 
 	rv.chunkSize = chunkSize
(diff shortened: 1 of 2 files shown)


abhinavdangeti (commented):
@moshaad7 per our conversation, let's get rid of the redundant check in posting.go now that you've handled it within this method.

Let's also get some clarity on what happens when we return an `error` when we stumble into this situation.

abhinavdangeti (changes requested):
Per previous comment.

abhinavdangeti:
@moshaad7 check out the latest 2 stack traces shared here - https://github.com/blevesearch/bleve/issues/1662#issuecomment-2353021096 .

It seems even if we handle the error case of chunkSize being zero prior to setting posting's attribute, you could run into divide by zero panics downstream on account of not updating the attribute in line 644 and 747 of posting.go which we'll need to handle as well.

posting.go
```diff
@@ -632,6 +629,10 @@ func (i *PostingsIterator) nextDocNumAtOrAfter(atOrAfter uint64) (uint64, bool,
 		return i.nextDocNumAtOrAfterClean(atOrAfter)
 	}
 
+	if i.postings.chunkSize == 0 {
+		return 0, false, fmt.Errorf("chunkSize is 0")
```
abhinavdangeti, line 633: Shouldn't you be returning `ErrChunkSizeZero` here?

posting.go
```diff
@@ -741,6 +742,10 @@ func (i *PostingsIterator) nextDocNumAtOrAfterClean(
 		return uint64(i.Actual.Next()), true, nil
 	}
 
+	if i.postings != nil && i.postings.chunkSize == 0 {
+		return 0, false, fmt.Errorf("chunkSize is 0")
```
abhinavdangeti, line 746: Ditto

abhinavdangeti (changes requested):
@moshaad7 one question still stands though - if we report this error now (better than the panic of course), how is the situation dealt with by the caller(s)?

abhinavdangeti (approved):

abhinavdangeti:
I'd like for one more review here from the team.

CascadingRadium (approved):

PR #267, opened 2024-10-03
Bug fixes and Optimizations

CascadingRadium:
- Fix issue where `vecIDsToExclude` is always nil if the cache entry exists, allowing deleted documents to be returned.
- Optimize vector cache logic by checking for a cache entry after acquiring a write lock to reduce redundant work during concurrent access.
- Resolve issue of duplicate entries being added to `dictLocs` when constructing an in-memory segment.
- Clean up code by removing the unused method parameters that were being passed around.

### DIFF
--- a/build.go
+++ b/build.go
@@ -163,16 +163,18 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32, numDocs uint64
 		mem:                 mem,
 		memCRC:              memCRC,
 		chunkMode:           chunkMode,
-		fieldsMap:           make(map[string]uint16),
 		numDocs:             numDocs,
 		storedIndexOffset:   storedIndexOffset,
 		fieldsIndexOffset:   sectionsIndexOffset,
 		sectionsIndexOffset: sectionsIndexOffset,
 		fieldDvReaders:      make([]map[uint16]*docValueReader, len(segmentSections)),
 		docValueOffset:      0, // docValueOffsets identified automatically by the section
-		dictLocs:            make([]uint64, 0),
 		fieldFSTs:           make(map[uint16]*vellum.FST),
 		vecIndexCache:       newVectorIndexCache(),
+		// following fields gets populated by loadFieldsNew
+		fieldsMap: make(map[string]uint16),
+		dictLocs:  make([]uint64, 0),
+		fieldsInv: make([]string, 0),
 	}
 	sb.updateSize()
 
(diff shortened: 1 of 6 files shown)


Likith101 (dismissed):

build.go
```diff
@@ -157,22 +157,20 @@ func persistStoredFieldValues(fieldID int,
...
-	storedIndexOffset uint64, dictLocs []uint64,
-	sectionsIndexOffset uint64) (*SegmentBase, error) {
+func InitSegmentBase(mem []byte, memCRC uint32, chunkMode uint32, numDocs uint64,
+	storedIndexOffset uint64, sectionsIndexOffset uint64) (*SegmentBase, error) {
 	sb := &SegmentBase{
 		mem:                 mem,
 		memCRC:              memCRC,
 		chunkMode:           chunkMode,
-		fieldsMap:           fieldsMap,
+		fieldsMap:           make(map[string]uint16),
```
abhinavdangeti, line 166: Can you add a comment above on where `fieldsMap` and `dictLocs` are _actually_ being updated.

abhinavdangeti (changes requested):

merge.go
```diff
@@ -111,7 +111,7 @@ func mergeSegmentBases(segmentBases []*SegmentBase, drops []*roaring.Bitmap, pat
 
 func MergeToWriter(segments []*SegmentBase, drops []*roaring.Bitmap,
```
abhinavdangeti, line 112: Does this API need to be exported?

CascadingRadium: no, it does not find its usage in bleve
i will unexport it

build.go
CascadingRadium: ok

segment.go
```diff
@@ -867,15 +867,6 @@ func (s *SegmentBase) loadDvReaders() error {
...
 
-				dataLoc, n := binary.Uvarint(s.mem[pos : pos+binary.MaxVarintLen64])
-				if n <= 0 {
-					return fmt.Errorf("loadDvReaders: failed to read the dataLoc "+
-						"offset for sectionID %v field %v", secID, s.fieldsInv[fieldID])
-				}
-				if secID == SectionInvertedTextIndex {
-					s.dictLocs = append(s.dictLocs, dataLoc)
-					s.incrementBytesRead(uint64(n))
-				}
```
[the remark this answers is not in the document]
abhinavdangeti: It seems it's duplicate data as mentioned in the commit description 👍🏼

abhinavdangeti (approved):

PR #268, opened 2024-10-10
 Add Thesaurus API and Synonym Index Handling in Search

CascadingRadium:
- Add Thesaurus API to find equivalent terms for a given term.
- Enable Synonym Document objects with Synonym Field objects to provide Synonym Definitions.
  for creating the thesaurus in the search index.
- Add a synonym section to handle synonym document processing; persist the synonym index in 
  segments (separating it from the inverted and vector indexes), and manage the synonym index 
  merging during segment merges.
- Add command line tooling to access the thesaurus by parsing the segment file.
- Update zap.md to reflect the index file format for thesaurus support.


abhinavdangeti (changes requested):

section_synonym.go
```diff
@@ -0,0 +1,740 @@
...
+	"github.com/RoaringBitmap/roaring"
+	"github.com/RoaringBitmap/roaring/roaring64"
+	index "github.com/blevesearch/bleve_index_api"
+	seg "github.com/blevesearch/scorch_segment_api/v2"
+	"github.com/blevesearch/vellum"
+)
+
+func init() {
+	registerSegmentSection(SectionSynonymIndex, &synonymIndexSection{})
+	invertedIndexExclusionChecks = append(invertedIndexExclusionChecks, func(field index.Field) bool {
```
abhinavdangeti, line 34: Need to check if these exclusion checks are needed for the vector section as well - if yes/no, this seems quite a brittle way to support this going forward as we add more sections and we must come up with something better.

section_synonym.go
```diff
@@ -0,0 +1,740 @@
...
+
+	builder    *vellum.Builder
+	builderBuf bytes.Buffer
+
+	tmp0 []byte
+
+	thesaurusAddrs map[int]int
+}
+
+func (so *synonymIndexOpaque) Set(key string, value interface{}) {
```
abhinavdangeti, line 95: The exported APIs at the very least require code commentary, some code commentary around the others is a plus too.

segment.go
```diff
@@ -472,6 +478,47 @@ func (sb *SegmentBase) dictionary(field string) (rv *Dictionary, err error) {
...
+// Thesaurus returns the thesaurus with the specified name, or an empty thesaurus if not found.
+func (s *SegmentBase) Thesaurus(name string) (segment.Thesaurus, error) {
+	thesaurus, err := s.thesaurus(name)
+	if err == nil && thesaurus == nil {
+		return emptyThesaurus, nil
+	}
+	return thesaurus, err
+}
+
+func (sb *SegmentBase) thesaurus(name string) (rv *Thesaurus, err error) {
```
abhinavdangeti, line 490: If this is called no where else than above, drop the separate method.

segment.go
```diff
@@ -719,6 +767,25 @@ func (s *Segment) DictAddr(field string) (uint64, error) {
 	return s.dictLocs[fieldIDPlus1-1], nil
 }
 
+// ThesAddr is a helper function to compute the file offset where the
+// thesaurus is stored with the specified name.
+func (s *Segment) ThesAddr(name string) (uint64, error) {
```
abhinavdangeti, line 772: Rename this to the entire thing.

abhinavdangeti (changes requested):
A few more comments.

segment.go
CascadingRadium: ok

segment.go
CascadingRadium: done

section_synonym.go
CascadingRadium: ok

segment.go
CascadingRadium: The `thesaurus` method is used in `mergeAndPersistSynonymSection` as it returns the type directly, the `Thesaurus` method returns an interface and is used by Bleve/Scorch... The same 2 methods are there for Dictionary as well. 
https://github.com/blevesearch/zapx/blob/cdad5b4dec7be6980c5bb6866cc9ba6b3ed552de/segment.go#L430-L479

section_synonym.go
CascadingRadium: I don't think an `exclusion` check is required for the vector/synonym sections ...
since they have `inclusion` checks each. The invertedTextIndex is a `catch-all` section that must process every field unless excluded (similar to default in switch-case). Moving forward, any new section will have to specify a special field type that only it can process.

The vector section can only process index.VectorField:

https://github.com/blevesearch/zapx/blob/3609ffc5f71b1d99026454ba60bf563f16cbe6ac/section_faiss_vector_index.go#L54-L57

The synonym section can only process index.SynonymField:
https://github.com/blevesearch/zapx/blob/3609ffc5f71b1d99026454ba60bf563f16cbe6ac/section_synonym.go#L387-L390

section_inverted_text_index.go
```diff
@@ -34,16 +34,27 @@ func init() {
...
 
-// this function is something that tells the inverted index section whether to
-// process a particular field or not - since it might be processed by another
-// section this function helps in avoiding unnecessary work.
-// (only used by faiss vector section currently, will need a separate API for every
-// section we introduce in the future or a better way forward - TODO)
-var isFieldNotApplicableToInvertedTextSection func(field index.Field) bool
+// This function checks whether the inverted index section should process
+// a particular field, avoiding unnecessary work if another section will handle it.
+var isFieldExcludedFromInvertedIndex = func(field index.Field) bool {
```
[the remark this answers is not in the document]
CascadingRadium: not required for `faissVectorIndexSection`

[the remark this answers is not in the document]
abhinavdangeti: Can you expand why?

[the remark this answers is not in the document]
abhinavdangeti: I see you've commented on this below.

section_synonym.go
abhinavdangeti: Ok, then would you add a NOTE to where we defined these exclusion checks that why this is applicable _only_ to InvertedTextIndexSection.

CascadingRadium: done

abhinavdangeti (commented):
Looks mostly ok to me. One comment around naming/commentary.

abhinavdangeti (approved):
Thanks @CascadingRadium . Let's get this in for now and incrementally improve the area as and when needed.

CascadingRadium:
Thanks for merging @abhinavdangeti

PR #269, opened 2024-10-16
MB-62230 - Pre-filtering performance optimisations

metonymic-smokey:
1. Add elements to the bitmap in a batch instead of individually to bring down latencies of pre-filtered queries.
2. Vary the selector used and use an exclude selector with high selectivity.

### DIFF
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -416,6 +416,13 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 
 					// Getting the vector IDs corresponding to the eligible
 					// doc IDs.
+					// The docVecIDMap maps each docID to vectorIDs corresponding
+					// to it.
+					// Usually, each docID has one vecID mapped to it unless
+					// the vector is nested, in which case there can be multiple
+					// vectorIDs mapped to the same docID.
+					// Eg. docID d1 -> vecID v1, for the first case
+					// d1 -> {v1,v2}, for the second case.
 					eligibleVecIDsBitmap := roaring.NewBitmap()
 					vecIDsUint32 := make([]uint32, 0)
 					for _, eligibleDocID := range eligibleDocIDs {
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -380,6 +380,19 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 				if len(eligibleDocIDs) > 0 {
 					// Non-zero documents eligible per the filter query.
 
+					// If every element in the index is eligible(eg. high selectivity
+					// cases), then this can basically be considered unfiltered kNN.
+					if len(eligibleDocIDs) == int(sb.numDocs) {
+						scores, ids, err := vecIndex.SearchWithoutIDs(qVector, k,
+							vectorIDsToExclude, params)
+						if err != nil {
+							return nil, err
+						}
+
+						addIDsToPostingsList(rv, ids, scores)
+						return rv, nil
+					}
+
 					// vector IDs corresponding to the local doc numbers to be
 					// considered for the search
 					vectorIDsToInclude := make([]int64, 0, len(eligibleDocIDs))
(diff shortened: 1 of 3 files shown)


CascadingRadium (approved):

faiss_vector_posting.go
```diff
@@ -417,20 +417,45 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 					// Getting the vector IDs corresponding to the eligible
 					// doc IDs.
 					eligibleVecIDsBitmap := roaring.NewBitmap()
+					vecIDsUint32 := make([]uint32, 0)
```
abhinavdangeti, line 420: Hmm, couple things ..

- does docVecIDMap need to of type map[uint32][]int64 and not map[uint32][]uint32?
- maybe better to not allocate here and do this instead .. (because `append` below in your code will become an alloc too)

```
eligibleVecIDsBitmap := roaring.NewBitmap()
for _, eligibleDocID := range eligibleDocIDs {
    vecIDs := docVecIDMap[uint32(eligibleDocID)]
    if len(vecIDs) > 0 {
        vecIDsUint32 := make([]uint32, len(vecIDs)
        for _, vecID := range vecIDs {
             vecIDsUint32 = append(vecIDsUint32, uint32(vecID))
        }
        eligibleVecIDsBitmap.AddMany(vecIDsUint32)
}
```

faiss_vector_posting.go
```diff
@@ -417,20 +417,45 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
 						for _, vecID := range vecIDs {
-							eligibleVecIDsBitmap.Add(uint32(vecID))
+							vecIDsUint32 = append(vecIDsUint32, uint32(vecID))
 						}
 					}
+					eligibleVecIDsBitmap.AddMany(vecIDsUint32)
+
+					var selector faiss.Selector
+					var err error
+					ineligibleVectorIDs := make([]int64, 0, len(vecDocIDMap)-
```
abhinavdangeti, line 431: Move this allocation to inside the `if` at line 433.

faiss_vector_posting.go
```diff
@@ -417,20 +417,45 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
+					}
 
 					// Determining which clusters, identified by centroid ID,
 					// have at least one eligible vector and hence, ought to be
 					// probed.
 					eligibleCentroidIDs := make([]int64, 0)
 					for centroidID, vecIDs := range centroidVecIDMap {
 						vecIDs.And(eligibleVecIDsBitmap)
-						if vecIDs.GetCardinality() > 0 {
+						if !vecIDs.IsEmpty() {
```
abhinavdangeti, line 458: 👍🏼

abhinavdangeti (changes requested):

faiss_vector_posting.go
```diff
@@ -417,20 +417,45 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
 						}
 					}
+					eligibleVecIDsBitmap.AddMany(vecIDsUint32)
+
+					var selector faiss.Selector
+					var err error
+					ineligibleVectorIDs := make([]int64, 0, len(vecDocIDMap)-
+						len(vectorIDsToInclude))
+					if float32(eligibleVecIDsBitmap.GetCardinality())/
+						float32(len(vecDocIDMap)) > 0.5 {
```
abhinavdangeti, line 434: Throw in a comment here on 0.5 also will you - as obvious as it may be :)

faiss_vector_posting.go
metonymic-smokey: :)

faiss_vector_posting.go
metonymic-smokey: About the first question, it's int64 since the vector IDs are of type int64.
I discussed this with @Thejas-bhat yesterday and he said that the rationale behind originally picking int64, over int32, was to minimise the chance of hash collisions due to overflow when computing the hash.

metonymic-smokey: @Thejas-bhat pls let me know if I got that right.

Thejas-bhat: yeah, makes sense to me 

another point here is that faiss accepts the ids as int64 type from what i see

https://github.com/blevesearch/faiss/blob/b747c55a93a9627039c34d44b081f375dca94e57/faiss/MetricType.h#L39
https://github.com/blevesearch/faiss/blob/b747c55a93a9627039c34d44b081f375dca94e57/faiss/Index.h#L116

metonymic-smokey: About the second point, I think your approach might have the downside of adding multiple batches to the eligibleVecIDsBitmap, instead of the current approach which adds a single one.

abhinavdangeti: Roaring is a much more efficient data structure, with the current approach - we'd be building one large uint32 array as opposed to multiple smaller ones - for GO GC, would the former be better or latter?

abhinavdangeti: Per clarification from Aditi - it seems the []int64 of vector IDs in the docVecIDMap is usually of length 1, except in the event of nested vectors, in which event the current implementation should be good.

@metonymic-smokey would you add some commentary around this here, also perhaps a couple examples showing how the docVecIDMap looks typically with and without nested vectors in those comments would be greatly helpful 👍🏼

abhinavdangeti (approved):

PR #275, opened 2024-10-21
MB-62230 - More pre-filtering optimisations

metonymic-smokey:
Use AddMany() when populating cluster assignment map with vectors belonging to a cluster.

### DIFF
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -425,9 +425,11 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 						if _, exists := centroidVecIDMap[centroidID]; !exists {
 							centroidVecIDMap[centroidID] = roaring.NewBitmap()
 						}
+						vecIDsUint32 := make([]uint32, 0, len(vecIDs))
 						for _, vecID := range vecIDs {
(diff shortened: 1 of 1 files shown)


CascadingRadium (approved):

faiss_vector_posting.go
```diff
@@ -425,9 +425,11 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 						if _, exists := centroidVecIDMap[centroidID]; !exists {
 							centroidVecIDMap[centroidID] = roaring.NewBitmap()
 						}
+						vecIDsUint32 := make([]uint32, 0, len(vecIDs))
 						for _, vecID := range vecIDs {
-							centroidVecIDMap[centroidID].Add(uint32(vecID))
```
metonymic-smokey, line 429: Consider: since the cluster assignment for an index is fixed, cache it like we do with the vecDocID map, if this PR shows some improvement in perf tests?

CascadingRadium: yes good idea

abhinavdangeti: Have you confirmed this to show improvements in perf at the unit level? (considering that the last set of changes that had similar improvements also didn't show any difference at all).

abhinavdangeti (approved):

PR #276, opened 2024-10-24
MB-63992 - High selectivity specific optimisations.

metonymic-smokey:
This PR adds minimises adding vector IDs to a bitmap, even in batches, for high selectivity cases.
It instead adds ineligible vector IDs to the bitmap.

### DIFF
--- a/faiss_vector_posting.go
+++ b/faiss_vector_posting.go
@@ -432,70 +432,86 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 						centroidVecIDMap[centroidID].AddMany(vecIDsUint32)
 					}
 
-					// Getting the vector IDs corresponding to the eligible
-					// doc IDs.
-					// The docVecIDMap maps each docID to vectorIDs corresponding
-					// to it.
-					// Usually, each docID has one vecID mapped to it unless
-					// the vector is nested, in which case there can be multiple
-					// vectorIDs mapped to the same docID.
-					// Eg. docID d1 -> vecID v1, for the first case
-					// d1 -> {v1,v2}, for the second case.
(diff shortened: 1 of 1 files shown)


abhinavdangeti:
@metonymic-smokey would you get this PR ready to go?

faiss_vector_posting.go
```diff
@@ -310,7 +310,7 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 	// Params needed for the closures
 	var vecIndex *faiss.IndexImpl
 	var vecDocIDMap map[int64]uint32
-	var docVecIDMap map[uint32][]int64
+	var docVecIDMap map[int64][]uint32
```
abhinavdangeti, line 313: Why the change in type for the key in this map?

metonymic-smokey: good point, it can remain uint32.

faiss_vector_posting.go
```diff
@@ -432,72 +432,88 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
-									if except != nil && !except.Contains(docID) {
-										ineligibleVectorIDs = append(ineligibleVectorIDs,
-											int64(vecID))
-									} else {
-										ineligibleVectorIDs = append(ineligibleVectorIDs,
-											int64(vecID))
-									}
+							if _, exists := eligibleDocIDsMap[uint64(docID)]; !exists {
+								for _, vecID := range vecIDs {
+									ineligibleVecIDsBitmap.Add(uint32(vecID))
```
abhinavdangeti, line 458: Didn't you want to replace this kind of a thing with `AddMany` instead?

abhinavdangeti: But that'd mean making another []uint32 array, so ok I guess.

PR #277, opened 2024-10-25
MB-57871 - [v15] Avoid unnecessary atomic operation on bytesWritten

Thejas-bhat:
- the construct of bytes written while constructing a segment is not a shared resource i.e. a single thread operates on a segment creation. Hence the atomic.Add() operation is not necessary here.
- Tested using bleve unit test with race flace 

```
go test -run=TestBytesWritten --race
PASS
ok  	github.com/blevesearch/bleve/v2	2.294s
```

### DIFF
--- a/contentcoder.go
+++ b/contentcoder.go
@@ -19,7 +19,6 @@ import (
 	"encoding/binary"
 	"io"
 	"reflect"
-	"sync/atomic"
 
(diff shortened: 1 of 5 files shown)


abhinavdangeti (approved):

abhinavdangeti:
Created `v15.3.16` to include this.

PR #282, opened 2024-11-28
MB-64360 - Guardrail fix for missing vectors

metonymic-smokey:


faiss_vector_posting.go
```diff
@@ -508,45 +510,49 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
+							selector, len(vectorIDsToInclude), closestCentroidIDs,
+							minEligibleCentroids, k, qVector, centroidDistances, params)
+						if err != nil {
+							return nil, err
+						}
 
-					addIDsToPostingsList(rv, ids, scores)
+						addIDsToPostingsList(rv, ids, scores)
+					}
 					return rv, nil
```
Thejas-bhat, line 556: coding practice/maintenance wise: i think the usual practice in situations such

```
var value 
if cond {
  // changes to value specific to cond
  return value
}
return value
```
its better to avoid to the return statement present in the condition block since it anyways falls to the return value outside block.

faiss_vector_posting.go
```diff
@@ -475,7 +475,9 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 							}
 						}
 
-						selector, err = faiss.NewIDSelectorNot(ineligibleVectorIDs)
+						if len(vectorIDsToInclude) > 0 {
```
Thejas-bhat, line 478: i think its better to short circuit these code paths to early exit the `rv` at 403 line and not have to do the downstream operations from thereon, since they're anyways not valid and we return an empty postings at 556.

metonymic-smokey: yeah, that makes it so much easier actually. great find!

Thejas-bhat (approved):

PR #317, opened 2025-03-25
MB-65473: Refactor and Optimize Pre-Filtered Vector Search

CascadingRadium:
- Refactor pre-filtered vector search to enhance performance and reduce memory footprint.
- Replace the current bitmap-based cluster selection mechanism with a simpler approach that uses the DirectMap in the IVF index. The IVF index's DirectMap directly maps the vector ID to the cluster it belongs to.
- Make `github.com/bits-and-blooms/bitset` a direct dependency of `zapx` and upgrade it to  
  the latest version
- Requires https://github.com/blevesearch/go-faiss/pull/41


faiss_vector_posting.go
```diff
@@ -395,156 +392,86 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
-					}
-					centroidVecIDMap[centroidID].AddMany(vecIDsUint32)
-				}
-
 				// Determining which clusters, identified by centroid ID,
 				// have at least one eligible vector and hence, ought to be
 				// probed.
-				eligibleCentroidIDs := make([]int64, 0)
-
+				clusterVectorCounts, err := vecIndex.CountVectorsPerCluster(vectorIDsToInclude)
```
abhinavdangeti, line 426: Rename this, tandem comment in https://github.com/blevesearch/go-faiss/pull/41#pullrequestreview-2714000662

abhinavdangeti (commented):
Just the rename request, looks fine otherwise.

faiss_vector_posting.go
```diff
@@ -395,156 +392,86 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
+					return nil, err
+				}
 				// Getting the nprobe value set at index time.
-				nprobe := vecIndex.GetNProbe()
-
-				eligibleDocsTillNow := int64(0)
-				minEligibleCentroids := 0
+				nprobe := int(vecIndex.GetNProbe())
+				var eligibleDocsTillNow int64
+				minEligibleCentroids := len(closestCentroidIDs) // Default to all centroids
```
metonymic-smokey, line 466: question about this decision: why did you decide to default to all centroids instead of the closest nprobe?
IMO defaulting to all centroids makes it similar to a flat index by default.

CascadingRadium: old code 
```
                eligibleDocsTillNow := int64(0)
				minEligibleCentroids := 0
				for i, centroidID := range closestCentroidIDs {
					eligibleDocsTillNow += int64(centroidVecIDMap[centroidID].GetCardinality())
					if eligibleDocsTillNow >= k && i >= int(nprobe-1) {
						// Continue till at least 'K' cumulative vectors are
						// collected or 'nprobe' clusters are examined, whichever
						// comes later.
						minEligibleCentroids = i + 1
						break
					}
					minEligibleCentroids = i + 1
				}
 ```
new code 

```
	            var eligibleDocsTillNow int64
				minEligibleCentroids := len(closestCentroidIDs) // Default to all centroids
				for i, centroidID := range closestCentroidIDs {
					eligibleDocsTillNow += clusterVectorCounts[centroidID]
					// Stop once we've examined at least 'nprobe' centroids and collected at least 'k' vectors.
					if eligibleDocsTillNow >= k && i+1 >= nprobe {
						minEligibleCentroids = i + 1
						break
					}
				}
```

the logic is same in both the code, 
in old code if loop terminates without ever satisfying the `if` condition, the final value of `minEligibleCentroids` will be equal to`  len(closestCentroidIds)`. In new code i just do it beforehand, in both cases when if condition is satisfied it will override the value with the correct value

**Comment by Thejas-bhat** [comment]
could you just update the PR description as to what's the change and how its supposed to give us benefits (jsut for my reference to start reviewing the code tomorrow)

File: faiss_vector_posting.go [hunk 3]
```diff
@@ -395,156 +392,102 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
 				}
-
 				// vector IDs corresponding to the local doc numbers to be
 				// considered for the search
 				vectorIDsToInclude := make([]int64, 0, len(eligibleDocIDs))
 				for _, id := range eligibleDocIDs {
-					vectorIDsToInclude = append(vectorIDsToInclude, docVecIDMap[uint32(id)]...)
+					vecIDs := docVecIDMap[uint32(id)]
+					if len(vecIDs) == 1 {
+						vectorIDsToInclude = append(vectorIDsToInclude, vecIDs[0])
```
Thejas-bhat, line 404: the if and else conditions aren't needed though right? since they're both same as `vectorIDsToInclude = append(vectorIDsToInclude, vecIDs...)`?

metonymic-smokey: LGTM other than this one.

CascadingRadium: This optimization is based on observations from CPU profiles.

In the most common case, the `docVecIDMap` maps each document to a single vector, meaning the vecIDs slice has only one element.
```
<doc1> -> [vec1]
<doc2> -> [vec2]
<doc3> -> [vec3]
```
With the original approach, an unnecessary overhead arises from the slice unpack-append operation, even though the typical case involves a single-element slice.

Consider the following Go code: 
```
package main

import (
	"math/rand"
	"testing"
)

var docVecIDMap map[uint32][]int64
var eligibleDocIDs []uint32

func init() {
	// Initialize test data
	docVecIDMap = make(map[uint32][]int64)
	eligibleDocIDs = make([]uint32, 10000)

	for i := 0; i < 10000; i++ {
		id := uint32(i)
		eligibleDocIDs[i] = id
		docVecIDMap[id] = []int64{rand.Int63()}
	}
}

func BenchmarkOriginalApproach(b *testing.B) {
	for i := 0; i < b.N; i++ {
		vectorIDsToInclude := make([]int64, 0, len(eligibleDocIDs))
		for _, id := range eligibleDocIDs {
			vectorIDsToInclude = append(vectorIDsToInclude, docVecIDMap[id]...)
		}
	}
}

func BenchmarkOptimizedApproach(b *testing.B) {
	for i := 0; i < b.N; i++ {
		vectorIDsToInclude := make([]int64, 0, len(eligibleDocIDs))
		for _, id := range eligibleDocIDs {
			vecIDs := docVecIDMap[id]
			if len(vecIDs) == 1 {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs[0])
			} else {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs...)
			}

		}
	}
}
```

The results of the benchmark test on my system are as follows:
- BenchmarkOriginalApproach
```
pkg: example.com/m
cpu: Apple M1 Pro
=== RUN   BenchmarkOriginalApproach
BenchmarkOriginalApproach
BenchmarkOriginalApproach-10                1004           1155589 ns/op           81920 B/op          1 allocs/op
PASS
ok      example.com/m   2.818s
```
- BenchmarkOptimizedApproach
```
pkg: example.com/m
cpu: Apple M1 Pro
=== RUN   BenchmarkOptimizedApproach
BenchmarkOptimizedApproach
BenchmarkOptimizedApproach-10               2007            563949 ns/op           81922 B/op          1 allocs/op
PASS
ok      example.com/m   2.443s
```

This indicates that the optimized approach is 2x faster.

CascadingRadium:
Profiles from high-selectivity tests indicated that a significant amount of time was spent generating ineligible vector IDs. This was primarily due to the overhead of hashing and handling hash collisions when using a map as a set, especially in large segments where the number of eligibleDocIDs is high.
```
	eligibleDocIDsSet := make(map[uint64]struct{}, len(eligibleDocIDs))
	for _, docID := range eligibleDocIDs {
		eligibleDocIDsSet[docID] = struct{}{}
	}
	ineligibleVectorIDs := make([]int64, 0, len(vecDocIDMap)-len(vectorIDsToInclude))
	for docID, vecIDs := range docVecIDMap {
		if _, exists := eligibleDocIDsSet[uint64(docID)]; !exists {
```
To address this, I replaced the map with a bitset from bits-and-blooms, an existing dependency of zapx. This optimization is particularly effective in high-selectivity cases, where at least 50% of the segment's docNums are eligible. Additionally, since the segment-local docNums increase monotonically from 0, the data in the bitset is dense rather than sparse. As a result, the bitset outperforms both the map and bitmap implementations for this use case.

Benchmark results confirm that the **bitset is the clear winner**, as expected.

Check the benchmark file below:

```
package main

import (
	"math/rand"
	"testing"

	"github.com/RoaringBitmap/roaring/v2/roaring64"
	"github.com/bits-and-blooms/bitset"
)

var docVecIDMap map[uint32][]int64
var vecDocIDMap map[int64]uint32
var eligibleDocIDs []uint64

func init() {
	// Initialize test data
	// Total documents in segment = 100K
	totalDocs := 100000
	// Eligible docs = 80K (80% selectivity)
	eligibleDocs := 80000
	// Test Data structures
	docVecIDMap = make(map[uint32][]int64, totalDocs)
	vecDocIDMap = make(map[int64]uint32, totalDocs)
	eligibleDocIDs = make([]uint64, eligibleDocs)
	// Populate DocVecIDMap with random data
	for i := 0; i < totalDocs; i++ {
		docID := uint32(i)
		vecID := int64(rand.Int63())
		docVecIDMap[docID] = []int64{vecID}
		vecDocIDMap[vecID] = docID
	}
	// Populate eligibleDocIDs
	for i := 0; i < eligibleDocs; i++ {
		id := uint64(i)
		eligibleDocIDs[i] = id
	}
	_ = float32(len(eligibleDocIDs)) / float32(len(docVecIDMap)) // 0.8 which is > 0.5
}

func BenchmarkMapSet(b *testing.B) {
	for i := 0; i < b.N; i++ {
		vectorIDsToInclude := make([]int64, 0, len(eligibleDocIDs))
		for _, id := range eligibleDocIDs {
			vecIDs := docVecIDMap[uint32(id)]
			if len(vecIDs) == 1 {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs[0])
			} else {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs...)
			}
		}
		eligibleDocIDsSet := make(map[uint64]struct{}, len(eligibleDocIDs))
		for _, docID := range eligibleDocIDs {
			eligibleDocIDsSet[docID] = struct{}{}
		}
		ineligibleVectorIDs := make([]int64, 0, len(vecDocIDMap)-len(vectorIDsToInclude))
		for docID, vecIDs := range docVecIDMap {
			if _, exists := eligibleDocIDsSet[uint64(docID)]; !exists {
				if len(vecIDs) == 1 {
					ineligibleVectorIDs = append(ineligibleVectorIDs, vecIDs[0])
				} else {
					ineligibleVectorIDs = append(ineligibleVectorIDs, vecIDs...)
				}
			}
		}
	}
}

func BenchmarkBitSet(b *testing.B) {
	for i := 0; i < b.N; i++ {
		vectorIDsToInclude := make([]int64, 0, len(eligibleDocIDs))
		for _, id := range eligibleDocIDs {
			vecIDs := docVecIDMap[uint32(id)]
			if len(vecIDs) == 1 {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs[0])
			} else {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs...)
			}
		}
		bs := bitset.New(uint(len(eligibleDocIDs)))
		for _, docID := range eligibleDocIDs {
			bs.Set(uint(docID))
		}
		ineligibleVectorIDs := make([]int64, 0, len(vecDocIDMap)-len(vectorIDsToInclude))
		for docID, vecIDs := range docVecIDMap {
			if !bs.Test(uint(docID)) {
				if len(vecIDs) == 1 {
					ineligibleVectorIDs = append(ineligibleVectorIDs, vecIDs[0])
				} else {
					ineligibleVectorIDs = append(ineligibleVectorIDs, vecIDs...)
				}
			}
		}
	}
}

func BenchmarkRoaringBitmap(b *testing.B) {
	for i := 0; i < b.N; i++ {
		vectorIDsToInclude := make([]int64, 0, len(eligibleDocIDs))
		for _, id := range eligibleDocIDs {
			vecIDs := docVecIDMap[uint32(id)]
			if len(vecIDs) == 1 {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs[0])
			} else {
				vectorIDsToInclude = append(vectorIDsToInclude, vecIDs...)
			}
		}
		bm := roaring64.NewBitmap()
		bm.AddMany(eligibleDocIDs)
		ineligibleVectorIDs := make([]int64, 0, len(vecDocIDMap)-len(vectorIDsToInclude))
		for docID, vecIDs := range docVecIDMap {
			if !bm.Contains(uint64(docID)) {
				if len(vecIDs) == 1 {
					ineligibleVectorIDs = append(ineligibleVectorIDs, vecIDs[0])
				} else {
					ineligibleVectorIDs = append(ineligibleVectorIDs, vecIDs...)
				}
			}
		}
	}
}
```

Output:

```
rahulrampure@HP:~/Downloads/99 Selec$ go test -bench=. -benchtime=5s -benchmem
goos: darwin
goarch: arm64
pkg: example.com/m
cpu: Apple M1 Pro
BenchmarkMapSet-10           	     859	   6604175 ns/op	 2221768 B/op	      16 allocs/op
BenchmarkBitSet-10           	    2191	   2746060 ns/op	  821280 B/op	       4 allocs/op
BenchmarkRoaringBitmap-10    	    1526	   3894757 ns/op	 1205290 B/op	      45 allocs/op
PASS
ok  	example.com/m	19.256s
```

faiss_vector_posting.go
Thejas-bhat: oh right - makes sense, thanks.

faiss_vector_posting.go
```diff
@@ -395,168 +393,112 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
-						len(vectorIDsToInclude))
-
+					ineligibleVectorIDs := make([]int64, 0, len(vecDocIDMap)-len(vectorIDsToInclude))
 					for docID, vecIDs := range docVecIDMap {
-						if _, exists := eligibleDocIDsMap[uint64(docID)]; !exists {
-							for _, vecID := range vecIDs {
-								ineligibleVecIDsBitmap.Add(uint32(vecID))
-								ineligibleVectorIDs = append(ineligibleVectorIDs, vecID)
+						if !bs.Test(uint(docID)) {
+							if len(vecIDs) == 1 {
```
abhinavdangeti, line 448: I see the reason why you're doing this per the conversation with @Thejas-bhat above - but given how this looks I feel someone in the future may refactor this to simply appending the slice. So, would you add some commentary here and above on why we're doing the len(..) == 1 check and the else.

faiss_vector_posting.go
abhinavdangeti: Let's add some commentary here reflecting why this maneuver - same concern as below.

metonymic-smokey (approved):

abhinavdangeti (approved):

PR #318, opened 2025-03-26
MB-57888: Copying updated fields while merging to avoid race conditions

Likith101:
The maps cannot be updated in any way while the index is active. This is to avoid any race conditions.

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -634,15 +634,17 @@ func mergeFields(segments []*SegmentBase) (bool, []string) {
 
 // Combine updateFieldInfo from all segments
 func mergeUpdatedFields(segments []*SegmentBase) map[string]*index.UpdateFieldInfo {
-	var fieldInfo map[string]*index.UpdateFieldInfo
+	fieldInfo := make(map[string]*index.UpdateFieldInfo)
 
 	for _, segment := range segments {
 		for field, info := range segment.updatedFields {
-			if fieldInfo == nil {
-				fieldInfo = make(map[string]*index.UpdateFieldInfo)
-			}
 			if _, ok := fieldInfo[field]; !ok {
-				fieldInfo[field] = info
+				fieldInfo[field] = &index.UpdateFieldInfo{
+					Deleted:   info.Deleted,
+					Index:     info.Index,
+					Store:     info.Store,
+					DocValues: info.DocValues,
+				}
 			} else {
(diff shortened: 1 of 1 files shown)


abhinavdangeti (approved):

merge.go
```diff
@@ -634,15 +634,17 @@ func mergeFields(segments []*SegmentBase) (bool, []string) {
...
-			if fieldInfo == nil {
-				fieldInfo = make(map[string]*index.UpdateFieldInfo)
-			}
 			if _, ok := fieldInfo[field]; !ok {
-				fieldInfo[field] = info
+				fieldInfo[field] = &index.UpdateFieldInfo{
+					Deleted:   info.Deleted,
+					Index:     info.Index,
+					Store:     info.Store,
+					DocValues: info.DocValues,
```
abhinavdangeti, line 646: This feels quite brittle to be honest - say we update the contents of `UpdateFieldInfo` in the future and we miss updating this copy.

The better way here would be to introduce a Copy member for UpdateFieldInfo that will responsible for this.

abhinavdangeti (changes requested):

merge.go
```diff
@@ -634,15 +634,17 @@ func mergeFields(segments []*SegmentBase) (bool, []string) {
...
+	fieldInfo := make(map[string]*index.UpdateFieldInfo)
 
 	for _, segment := range segments {
 		for field, info := range segment.updatedFields {
-			if fieldInfo == nil {
-				fieldInfo = make(map[string]*index.UpdateFieldInfo)
-			}
 			if _, ok := fieldInfo[field]; !ok {
-				fieldInfo[field] = info
+				fieldInfo[field] = &index.UpdateFieldInfo{
```
CascadingRadium, line 642: keep the make inside and keep L637 as is, to avoid unnecessary map creation in the general case when no index update is done

```
if fieldInfo == nil {
    fieldInfo = make(map[string]*index.UpdateFieldInfo)
}
fieldInfo[field] = &index.UpdateFieldInfo{
	Deleted:   info.Deleted,
	Index:     info.Index,
	Store:     info.Store,
	DocValues: info.DocValues,
}
```

abhinavdangeti:
Reverting the base change with https://github.com/blevesearch/zapx/pull/319 ,  will need a fresh pull request (can be included in this PR).

PR #320, opened 2025-04-07
MB-65473: [BP] Refactor and Optimize Pre-Filtered Vector Search (#317)

CascadingRadium:
- Refactor pre-filtered vector search to enhance performance and reduce memory footprint.
- Replace the current bitmap-based cluster selection mechanism with a simpler approach that uses the DirectMap in the IVF index. The IVF index's DirectMap directly maps the vector ID to the cluster it belongs to.
- Make `github.com/bits-and-blooms/bitset` a direct dependency of `zapx` and upgrade it to
  the latest version
- Requires https://github.com/blevesearch/go-faiss/pull/41

---------

### DIFF
--- a/faiss_vector_cache.go
+++ b/faiss_vector_cache.go
@@ -105,7 +105,7 @@ func (vc *vectorIndexCache) addDocVecIDMapToCacheLOCKED(ce *cacheEntry) map[uint
 		return ce.docVecIDMap
 	}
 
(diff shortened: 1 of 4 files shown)


abhinavdangeti (approved):

abhinavdangeti:
Dammit, should've merged the index_api commits and updated go.mod first, let's raise a separate PR for the go.mod updates

PR #329, opened 2025-05-09
MB-62985 - Add support for binary quantised vectors. 

metonymic-smokey:


faiss_vector_cache.go
```diff
@@ -161,21 +163,39 @@ func (vc *vectorIndexCache) createAndCacheLOCKED(fieldID uint16, mem []byte,
 		}
 	}
 
+	indexes = make([]*faiss.IndexImpl, 0)
+	binaryIndexSize, n := binary.Uvarint(mem[pos : pos+binary.MaxVarintLen64])
```
CascadingRadium, line 167: this will break existing index having a plain vector index as the file format will be v16. 
Since this feature entails a file format change, would recommend you create a new v17 file format, copy all the code from v16 into it (text, synonyms) and then include your change there. 
Also as suggested in the doc, add a num_vector_index as a separate entry in the file format, so the file would look like so:
```
|num index|bin index|float index| -> if num index = 2
|num index|float index|           -> if num index = 1 
```
Also

faiss_vector_cache.go
```diff
@@ -161,21 +163,39 @@ func (vc *vectorIndexCache) createAndCacheLOCKED(fieldID uint16, mem []byte,
...
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
+	indexes = append(indexes, index)
 
-	vc.insertLOCKED(fieldID, index, vecDocIDMap, loadDocVecIDMap, docVecIDMap)
-	return index, vecDocIDMap, docVecIDMap, vecIDsToExclude, nil
+	cacheEntryStub := cacheEntryReqs{
+		index:       indexes[1],
+		binaryIndex: indexes[0],
```
CascadingRadium, line 190: maybe you can switch the order as well, for easier extensibility and simplifying upstream code. 
```
indexes[0] = float index
indexes[1] = bin index
indexes[2] = future use case
```

faiss_vector_cache.go
```diff
@@ -283,19 +302,33 @@ func (e *ewma) add(val uint64) {
 
 // -----------------------------------------------------------------------------
 
-func createCacheEntry(index *faiss.IndexImpl, vecDocIDMap map[int64]uint32,
-	loadDocVecIDMap bool, docVecIDMap map[uint32][]int64, alpha float64) *cacheEntry {
+// required info to create a cache entry.
+type cacheEntryReqs struct {
+	alpha       float64
+	index       *faiss.IndexImpl
```
CascadingRadium, line 308: maybe just have the indexes array itself here 

```
indexes []*faiss.IndexImpl
```

faiss_vector_cache.go
```diff
@@ -161,21 +163,39 @@ func (vc *vectorIndexCache) createAndCacheLOCKED(fieldID uint16, mem []byte,
...
+	}
+
+	vc.insertLOCKED(fieldID, cacheEntryStub)
+	return indexes, vecDocIDMap, docVecIDMap, vecIDsToExclude, nil
 }
 
-func (vc *vectorIndexCache) insertLOCKED(fieldIDPlus1 uint16,
-	index *faiss.IndexImpl, vecDocIDMap map[int64]uint32, loadDocVecIDMap bool,
-	docVecIDMap map[uint32][]int64) {
+func (vc *vectorIndexCache) insertLOCKED(fieldIDPlus1 uint16, ce cacheEntryReqs) {
```
CascadingRadium, line 198: take pointer to the cacheEntryReqs struct, don't copy the struct

faiss_vector_posting.go
```diff
@@ -354,12 +381,39 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
 					return rv, nil
 				}
 
-				scores, ids, err := vecIndex.SearchWithoutIDs(qVector, k,
-					vectorIDsToExclude, params)
+				binaryQueryVector := convertToBinary(qVector)
```
CascadingRadium, line 384: shouldn't you be doing this only when the vector field has been enabled as binary i.e

```
index_as_binary": true
```

faiss_vector_posting.go
```diff
@@ -354,12 +381,39 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
 
+				distances := make([]float32, k*4)
+				err = vecIndex.DistCompute(qVector, binIDs, int(k*4), distances)
+				if err != nil {
+					return nil, err
+				}
+
+				// Need to map distances to the original IDs to get the top K.
+				// Use a heap to keep track of the top K.
+				h := &maxHeap{}
```
CascadingRadium, line 399: I don't think this part is necessary, as the KNN Collector will handle this itself right.

The KNN collector picks top K across multiple segments using an optimised heap based approach.

So even if you give (where S is the number of segments)
- S * K (for the basic non binary vector search)
- S * K * 4 (for the binary vector search with oversampling factor 4)

the result will still be top K.
So having the overhead here to pick top K' from K * 4 and then picking final top K from from S * K' will be redundant.

faiss_vector_posting.go
metonymic-smokey: yep, that's the final plan, this is sorta a POC.
But I've had a couple of conversations with Abhi and we haven't yet decided on how we'll introduce this feature. We're leaning towards the possibility that this will be introduced as another index optimisation.

metonymic-smokey: Given that this doesn't make sense with the memory-efficient optimisation since it really tanks the recall to 40% from 56%, I think it's best to keep this as another optimisation.

faiss_vector_posting.go
metonymic-smokey: great point!

metonymic-smokey: just to double check, will run a recall test on the 500k docs with this change.

faiss_vector_cache.go
metonymic-smokey: i just thought of another point, I think there needs to be a 'type specifier' too for the indexes?
if there are future index types added, let's say types x and y, I think it would be useful to add _which_ index is stored in the section:
|num index| index_type_float| float index | index_type_x | x index |, for example.

CascadingRadium: This would help with code readability, but wouldn't it be redundant information? Since we already follow a fixed pattern when writing to the file, the same pattern has to be followed while reading it as well. So if we just document in the `zap.md` file that the format follows:
```
| num index | float index | bin index | X index | Y index |
```
then we can achieve the same clarity without needing an explicit type specifier.

Even if new index types like X or Y are added in the future, as long as we update the documentation to reflect the new structure, the reader will still know what to expect at each position.

metonymic-smokey: Let's say a user doesn't apply BQ optimizations, but only applies X and Y after the float index, won't it be clearer to read if we know that the pattern is float --> x --> y?

CascadingRadium: Oh yeah, i had missed that point. Then, yeah, the idea is good. Probably an `enum` can be used for this purpose similar to `vectorIndexOptimizedFor`

section_faiss_vector_index.go
```diff
@@ -578,6 +704,68 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
 			}
 		}
 
+		if content.indexOptimizedFor == index.IndexOptimizedForRecallBinary ||
```
CascadingRadium, line 707: wouldn't it be more idiomatic to have the binary index after the float index?

section_faiss_vector_index.go
```diff
@@ -578,6 +704,68 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
...
+			content.indexOptimizedFor == index.IndexOptimizedForLatencyBinary {
+			binaryIndexDescription := determineBinaryIndexToUse(nvecs, nlist)
+			binaryFaissIndex, err := faiss.IndexBinaryFactory(int(content.dim),
+				binaryIndexDescription, metric)
+			if err != nil {
+				return 0, err
+			}
+			defer binaryFaissIndex.Close()
+
+			bvecs := convertToBinary(vecs)
```
CascadingRadium, line 717: this looks like an issue. Assume we have 2 scenarios: 
- 4 vectors of dims 2 
- 2 vectors of dims 4
- 1 vector of dims 8

all three will be treated as one and the same by converting into 1 single binary vector (as you are passing `vecs` here which is of length `D * L`).
This would be wrong as we lose the posting data. Ideally if D is not a multiple of 8 we should not store any binary data at all.

section_faiss_vector_index.go
```diff
@@ -578,6 +704,68 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
 			}
 		}
 
+		if content.indexOptimizedFor == index.IndexOptimizedForRecallBinary ||
+			content.indexOptimizedFor == index.IndexOptimizedForLatencyBinary {
+			binaryIndexDescription := determineBinaryIndexToUse(nvecs, nlist)
+			binaryFaissIndex, err := faiss.IndexBinaryFactory(int(content.dim),
```
CascadingRadium, line 710: i dont think the IndexBinaryFactory can take a dims which is not divisible by 8

section_faiss_vector_index.go
```diff
@@ -578,6 +704,68 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) (offset uint
...
+				if err != nil {
+					return 0, err
+				}
+
+				faissIndex.SetNProbe(nprobe)
+
+				err = faissIndex.Train(vecs)
+				if err != nil {
+					return 0, err
+				}
```
CascadingRadium, line 730: this is already done in L639 to L650. Why are you training the regular faiss index twice on the same data? `vecs` is unchanged here

section_faiss_vector_index.go
```diff
@@ -476,6 +546,61 @@ func determineIndexToUse(nvecs, nlist int, indexOptimizedFor string) (string, in
...
+		// Apply the threshold: convert the float32 to 1 or 0 based on threshold
+		var bit uint8
+		if value >= 0.0 {
+			bit = 1
+		} else {
+			bit = 0
+		}
+
+		// Shift the bit into the correct position in the byte
+		currentByte |= (bit << (7 - bitCount))
```
CascadingRadium, line 577: `bit` seems redundant to me
isn't the code below equivalent?
```
if value >= 0.0 {
	currentByte |= (1 << (7 - bitCount))
}
```

section_faiss_vector_index.go
```diff
@@ -398,13 +413,68 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(sbs []*SegmentBase,
...
+		indexOptimizedFor == index.IndexOptimizedForLatencyBinary {
+		binaryIndexDescription := determineBinaryIndexToUse(nvecs, nlist)
+		binaryFaissIndex, err := faiss.IndexBinaryFactory(int(dims),
+			binaryIndexDescription, metric)
+		if err != nil {
+			return err
+		}
+		defer binaryFaissIndex.Close()
+
+		bvecs := convertToBinary(indexData)
```
CascadingRadium, line 434: Same issue as before. If D is a multiple of 8 this works. Else we will create a wrong index itself. (creating a binary index where multiple X-dimensional vectors from different documents get compressed to one vector, thereby losing the posting data)

section_faiss_vector_index.go
```diff
@@ -398,13 +413,68 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(sbs []*SegmentBase,
...
+			err = faissIndex.SetDirectMap(2)
+			if err != nil {
+				return err
+			}
+
+			faissIndex.SetNProbe(nprobe)
+
+			err = faissIndex.Train(indexData)
+			if err != nil {
+				return err
```
CascadingRadium, line 446: again not required as its already done above

faiss_vector_posting.go
```diff
@@ -354,14 +382,49 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, requiresFiltering bool
...
+					}
+
+					// Pop the top K in reverse order to get them in ascending order
+					ids := make([]int64, k)
+					scores := make([]float32, k)
+					for i := int(k) - 1; i >= 0; i-- {
+						distanceID := heap.Pop(h).(*distanceID)
+						scores[i] = distanceID.distance
+						ids[i] = distanceID.id
+					}
```
CascadingRadium, line 417: At this stage, we already have the complete set of ids and their corresponding distances available for Top-K selection. This scenario differs from the _K-sized heap approach used in `Bleve`_ (see [knn.go#L116](https://github.com/blevesearch/bleve/blob/e8d44258c631f775964785b1fb9e2719fdb8f76a/search/collector/knn.go#L116)), where the full result set `(N*K)` is not known ahead of time and must be streamed or iterated over(making the K-Sized Heap approach optimal).

In our case, since the entire dataset is available upfront, using a heap becomes unnecessarily expensive—particularly in terms of space complexity O(K). This overhead can compound significantly with the number of segments in a highly distributed index, as the heap becomes temporary query-specific state for each segment.

Given that, it would be more efficient to use a `Randomized QuickSelect` algorithm (https://courses.grainger.illinois.edu/cs473/sp2015/w/lec/15_notes.pdf)  for in-place Top-K selection. This approach reduces memory usage, avoids unnecessary allocations, and maintains excellent average-case performance (O(N)) suitable for batch-style processing where the full dataset is already materialized. Also this benefits the argment that we do not actually _need_ a sorted list of distances and ids, we just need the top-K in any order, as the upstream implementations already handle sorting the final result set based on scores

section_faiss_vector_index.go
metonymic-smokey: I think the opposite would be better since the binary index is queried before the float index, when used.

section_faiss_vector_index.go
metonymic-smokey: yep, I'm aware - did this for the PoC! Will fix this now that these PRs are shaping up for the final drop.

section_faiss_vector_index.go
CascadingRadium: ok, makes sense.

faiss_vector_posting.go
CascadingRadium: resolving as not applicable (due to performance regression - rel: quickselect comment) thanks

faiss_vector_posting.go
CascadingRadium: resolving as not applicable - design change to binary index optimization

faiss_vector_cache.go
CascadingRadium: resolving as done

section_faiss_vector_index.go
metonymic-smokey: sorry, my bad - this was an oversight.

PR #336, opened 2025-06-17
MB-62182: Fast merge of vector indexe

Thejas-bhat:
- Introduces a way to merge the vector indexes without having to re-train the merged index. 
- Uses the merge_from API from faiss which merges the underlying inverted lists (based on the identical centroid layout) in a block wise fashion.
- During the merge process, if the config tells us to train the vector index (during the initial part of index build before the data ingest) the vectors from the participating indexes are reconstructed + trained upon for the final merged index. If the config specifies us to use the "centroid index" to merge, we will now use that template for the merged index and also use the merge_from API to move all the vectors from the participating indexes to the final merged one.
- NOTE: for now, we're supporting the fast merge for purely doc addition type of workload - i.e. during the index build since that's the most common type of workload for vector databases. For the workload involving updates and deletes we end up falling back to naive merge approach.


abhinavdangeti:
@Thejas-bhat is this ready for review? Needs a rebase.

centroid_index.go
```diff
@@ -0,0 +1,74 @@
...
+	_, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+	pos += n
+	_, n = binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+	pos += n
+
+	// index optimization type
+	optType, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+	pos += n
+
+	if index.VectorIndexOptimizationsReverseLookup[int(optType)] != index.IndexOptimizedFastMerge {
```
abhinavdangeti: This stuff needs update, on account of the change in the approach.

faiss_vector_io_flags_win.go
```diff
@@ -20,3 +20,4 @@ package zap
 import faiss "github.com/blevesearch/go-faiss"
 
 const faissIOFlags = faiss.IOFlagReadOnly
+const faissIOFlagsReadOnly = faiss.IOFlagReadOnly
```
abhinavdangeti: Ditto.

faiss_vector_io_flags_unix.go
```diff
@@ -20,3 +20,4 @@ package zap
 import faiss "github.com/blevesearch/go-faiss"
 
 const faissIOFlags = faiss.IOFlagReadMmap | faiss.IOFlagSkipPrefetch
+const faissIOFlagsReadOnly = faiss.IOFlagReadOnly
```
abhinavdangeti: Why declare these constants - let's just re-use go-faiss's.

go.mod
```diff
@@ -20,3 +20,7 @@ require (
 	github.com/spf13/pflag v1.0.9 // indirect
 	golang.org/x/sys v0.40.0 // indirect
 )
+
+replace github.com/blevesearch/go-faiss => /Users/thejas.orkombu/fts/blevesearch/go-faiss
+
+replace github.com/blevesearch/scorch_segment_api/v2 => /Users/thejas.orkombu/fts/blevesearch/scorch_segment_api
```
abhinavdangeti: Undo replaces, instead use your branches and run go-mod-tidy.

section_faiss_vector_index.go
```diff
@@ -38,6 +38,9 @@ func init() {
 	faiss.SetOMPThreads(defaultFaissOMPThreads)
 }
 
+// the threshold of vector count above which SQ8 quantization is applied
+const sq8Threshold = 10000
```
abhinavdangeti: More of an IVF threshold than SQ8.

abhinavdangeti (changes requested):

section_faiss_vector_index.go
```diff
@@ -38,6 +38,9 @@ func init() {
 	faiss.SetOMPThreads(defaultFaissOMPThreads)
 }
 
+// the threshold of vector count above which SQ8 quantization is applied on ivf index
+const ivfSq8Threshold = 10000
```
abhinavdangeti, line 42: Let's just call this ivfThreshold

abhinavdangeti (commented):
Would you upgrade scorch_segment_api to v2.4.4.

section_faiss_vector_index.go
Thejas-bhat: but wouldn't ivfSq8 be more applicable since we also maintain the ivfflat indexes?

abhinavdangeti: Ah ok for now then.

abhinavdangeti: So fast merge is applicable only toward indexes larger than this threshold then, and reconstruct-merge will occur over the remaining?

abhinavdangeti (approved):

section_faiss_vector_index.go
Thejas-bhat: yeah that's right, like we're trying to cut down the cost on the large vector merge operations which happen in the ivfsq8 class. and also we wouldnt be able to copy the centroid layout from the trained index to the flat class since the training would have happened keeping this threshold in mind

Thejas-bhat:
let me run some more tests on my local system and then merge this change

centroid_index.go
```diff
@@ -0,0 +1,70 @@
...
+package zap
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	faiss "github.com/blevesearch/go-faiss"
+)
+
+func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
```
capemox, line 27: Why doesn't the function return a `faiss.indexImpl` directly?

Thejas-bhat: we're keeping the signature a generic interface{} so that we don't have to import the faiss package in bleve and scorch_segment_api. it also keeps the implementation hidden and the caller that's in the merge path will do an interface check (an acceptable cost so far what we've thought) and force it to IndexImpl before performing the merge.

abhinavdangeti:
`v17.0.6`

PR #351, opened 2025-12-11
MB-62985: Binary Quantization

Likith101:
- Added a new vector index type called binary index
 - Will be used alongside the regular float index
 - During regular search, we first run the query against the binary index and then only do the distance computing with the float index
 - Binary indexes are not merged, but are reconstructed


faiss_vector_wrapper.go
```diff
@@ -74,8 +82,15 @@ func (v *vectorIndexWrapper) Search(qVector []float32, k int64, params json.RawM
...
-	rs, err := v.searchWithoutIDs(qVector, k, v.exclude, params)
+
+	var rs resultSet
+	var err error
+
+	if v.bIndex != nil {
+		rs, err = v.searchBinaryWithoutIDs(qVector, k, v.exclude, params)
+	} else {
+		rs, err = v.searchFloatWithoutIDs(qVector, k, v.exclude, params)
+	}
```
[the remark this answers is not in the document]
CascadingRadium: Assuming we extend the code to support Index X (gpu index class for example) I will need to add new APIs which would lead to code duplication. 

Better to shift the index identification logic to the API itself. This ensures we do not duplicate the logic for the second callback for `docSearch`

```go

func (v *vectorIndexWrapper) searchWithoutIDs(qVector []float32, k int64, include *bitmap, params json.RawMessage) (
	resultSet, error) {
	// override the input params if binary for oversampling
	if index is Binary {
		qVector = convertToBinary(qVector, len(qVector))
		k = binaryOversampleValue * k
	}
	return v.docSearch(k, v.sb.numDocs,
		func() ([]float32, []int64, error) {
			// build the FAISS selector based on the exclude bitmap, if any.
			// The exclude bitmap can be nil, indicating no exclusions, in that
			// case we can pass a nil selector to FAISS.
			// NOTE: The bitmap selector is just a wrapper over the exclude bitmap
			// which is shared across the CGO layer.
			sel, err := getExcludeSelector(exclude)
			if err != nil {
				return nil, nil, err
			}
			// NOTE: the selector being freed does NOT free the inner bitmap, as we control
			// its lifecycle in GO, to reuse the bitmap across iterations, if needed, for
			// multi-vector document retrieval.
			if sel != nil {
				defer sel.Delete()
			}
			scores, labels, err := v.bIndex.SearchWithSelector(binQVector, binKVal, sel, params)
			if err != nil {
				return nil, nil, err
			}
			// need to consult float index to re calculate the real scores and labels, override the variables.
			if index is binary {
				distances, err = v.fIndex.DistCompute(qVector, binIDs)
				if err != nil {
					return nil, nil, err
				}
				scores, labels = TopNIDsByDistance(distances, binIDs, int(k))
			}
			return scores, labels
		},
		func(numIter int, labels []int64) bool {
			// if this is the first loop iteration and we have < k unique docIDs,
			// we must clone the existing exclude bitmap before modifying it
			// to avoid modifying the original bitmap passed in by the caller
			if numIter == 1 {
				// if we do not have an exclude bitmap yet, create a new one
				if exclude == nil {
					exclude = newBitmap(v.mapping.numVectors())
				} else {
					// clone the existing exclude bitmap
					exclude = exclude.clone()
				}
			}
			// prepare the exclude list for the next iteration by adding
			// the vector ids retrieved in this iteration
			for _, vecID := range labels {
				// should not happen, but just a safeguard, as we catch -1
				// in the main loop
				if vecID == -1 {
					continue
				}
				exclude.set(uint32(vecID))
			}
			// with exclude bitmap updated, we can proceed to the next iteration
			// fast check if the exclude bitmap has all vectors excluded, in which case
			// we can stop searching further
			return exclude.cardinality() != v.mapping.numVectors()
		})
}

File: section_faiss_vector_index.go [hunk 2]
```diff
@@ -578,8 +688,8 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) error {
 				return err
 			}
 		}
-		// write the type of the vector index (unused for now)
-		n = binary.PutUvarint(tempBuf, 0)
+		// write the type of the vector index
+		n = binary.PutUvarint(tempBuf, uint64(indexType))
```
CascadingRadium: since the `binaryIndex` is implied by the index optimization type = `binary` what is the purpose of persisting this additional information on disk?

CascadingRadium (changes requested):

abhinavdangeti (commented):
First pass

section_faiss_vector_index.go
```diff
@@ -55,6 +55,12 @@ const (
 	IndexTypeIVF
 )
 
+// Vector index section implementation types
+const (
+	FaissFP32Index uint64 = iota
+	FaissBIVFIndex
```
abhinavdangeti: I don't see the need for these to be exported?

abhinavdangeti (changes requested):
First pass

section_faiss_vector_index.go
Likith101: I believe it should be fine for now to generalise both bflat and bivf under the same name bivf. It also should not affect Rabitq if we choose to add it in the future

faiss_vector_wrapper.go
```diff
@@ -145,9 +156,21 @@ func (v *vectorIndexWrapper) SearchWithFilter(qVector []float32, k int64,
 	if numSelected == v.mapping.numVectors() {
 		return v.Search(qVector, k, params)
 	}
-	// If the index is not an IVF index, then the search can be
-	// performed directly, using the Flat index.
-	if !v.vecIndex.IsIVFIndex() {
+
+	// If a binary index is present and is not an IVF index,
+	// then we can perform the search directly on the binary index
+	if v.index.bIndex != nil && !v.index.bIndex.IsIVFIndex() {
```
CascadingRadium: this block is duplicated code. 
maybe u can have 

```go
func (idx *faissIndex) IsIVF() bool {
    if idx.bIndex != nil {
        return idx.bIndex.IsIVFIndex()
    }
    if idx.fIndex != nil {
        return idx.fIndex.IsIVFIndex()
    }
    return false
}
```
and then you can just 

```go
if !v.index.IsIVF() {
    rs, err := v.searchWithIDs(qVector, k, includeBM, params)
    if err != nil {
        return nil, err
    }
    return getPostingsList(rs), nil
}
```

faiss_vector_cache.go
```diff
@@ -299,15 +314,62 @@ func (ce *cacheEntry) load(except *roaring.Bitmap) (*faiss.IndexImpl, *idMapping
...
+			ce.index.close()
 		}
-		ce.index = nil
 		ce.mapping = nil
 	}()
 }
 
 // -----------------------------------------------------------------------------
 
+type faissIndex struct {
```
CascadingRadium: shift this util struct to faiss_vector_wrapper.go

section_faiss_vector_index.go
```diff
@@ -347,80 +360,233 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(sbs []*SegmentBase,
...
+		return err
+	}
+	// write the length of the serialized vector index bytes
+	n = binary.PutUvarint(tempBuf, uint64(len(fIndexBytes)))
+	_, err = w.Write(tempBuf[:n])
+	if err != nil {
+		return err
+	}
+	// write the vector index data
+	_, err = w.Write(fIndexBytes)
```
CascadingRadium:
```go
	if err != nil {
		return err
	}
```

faiss_vector_cache.go
Likith101: I feel like it makes more sense for it to be present along side the cache as it is the place where the structure is stored as opposed to the wrapper code which has all of the search related functions

CascadingRadium (dismissed):

capemox (dismissed):

abhinavdangeti (dismissed):
@Likith101 the overhead with a flat index (for the backing index) is going to unfortunately be a no-go. We need to figure out alternatives - either using RaBitQ without backing or IVF,BQ with a quantized index for backing.

CascadingRadium (approved):

CascadingRadium (dismissed):

capemox (dismissed):

capemox (approved):

CascadingRadium (approved):

maneuvertomars (approved):

PR #382, opened 2026-03-23
MB-71082: Add guard to avoid passing empty vector ID slice to selector

capemox:
If all vectors satisfy the prefilter criteria in `SearchWithFilter`, a selector would be created with a 0 length slice of ineligible vector IDs (since all vectors are eligible, no vectors will be ineligible). The selector tries to dereference this slice causing a panic.

We avoid this by adding a check to see if the eligible vectors list is the same as all the vectors in the index. If so, default to unfiltered search.

### DIFF
--- a/faiss_vector_test.go
+++ b/faiss_vector_test.go
@@ -5,6 +5,7 @@ package zap
 
 import (
 	"encoding/binary"
+	"fmt"
 	"math"
 	"math/rand"
 	"os"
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
@capemox thanks for this - could we add a unit test alongside? Also is this not applicable to `v17`?

capemox:
@abhinavdangeti yep, there's already a test in the PR! That test reproduces the panic scenario which fails without this change.

v17 actually already has this guard as part of the refactor: https://github.com/blevesearch/zapx/blob/385c20cc3a2ff3cadc6facde289bd165b14b2fe2/faiss_vector_wrapper.go#L156

This edge case was missed in v16

abhinavdangeti (approved):
Thanks @capemox  👍🏼

abhinavdangeti:
@capemox Let's propose this backport to `v16-7.6.x-couchbase` and `v16-8.0.x-couchbase` as well please.

abhinavdangeti:
`v16.3.2`

PR #384, opened 2026-03-24
[v17] Add Request Batcher Module

CascadingRadium:
- Add latency-aware batching for vector search requests using a timer-driven coalescing queue.
- Merge compatible requests (same k) and execute them together to reduce number of `Faiss` calls.
- Ensure bounded latency via per-request deadlines using a package-level variable called `LatencyBudget`.

### DIFF
--- a/faiss_vector_request_batcher.go
+++ b/faiss_vector_request_batcher.go
@@ -1,0 +1,260 @@
+//  Copyright (c) 2026 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build vectors
+// +build vectors
+
+package zap
+
+import (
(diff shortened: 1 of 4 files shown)


faiss_vector_request_batcher.go
```diff
@@ -0,0 +1,263 @@
...
+	// - Lower the latency budget, lower the throughput of the batcher, but also lower the latency for individual search requests.
+	LatencyBudget time.Duration = 250 * time.Millisecond
+)
+
+var (
+	errBatcherStopped error = errors.New("batcher has been stopped")
+)
+
+// The requestBatcher interface defines the contract for a component that batches search requests to a Faiss index.
+type requestBatcher interface {
```
capemox: No need to have this interface, can be a concrete struct

faiss_vector_request_batcher.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+}
+
+// --------------------------------------------------
+// coalesceQueue
+// --------------------------------------------------
+
+type coalesceQueue struct {
+	// the Faiss index that this coalesce queue will execute search requests against.
+	idx faissIndex
```
capemox: `idx` in coalesceQueue must be of type `faiss.GPUIndexImpl`. This is because the requestBatcher should be a part of the gpu impl of the faissIndex struct, and so request batcher will only be called if we go on the GPU path.

CascadingRadium: hey @capemox, this PR is meant to add an independent module for request batching. Making it specific for GPU usage would add coupling which is not something that i feel is wise.

CascadingRadium: done

capemox (approved):

PR #385, opened 2026-03-25
MB-59670: Add GPU support

capemox:
- New `faissGPUFloat32Index` implementing `faissIndex`. It opaquely performs operations on cpu or gpu when appropriate.
- Supports training and search on gpu, falls back to cpu when appropriate
- Reorganize `makeFaissIndex` to avoid losing direct map information when transferring to gpu


faiss_vector_cache.go
```diff
@@ -164,9 +165,20 @@ func (vc *vectorIndexCache) createAndCacheLOCKED(fieldID uint16, mem []byte,
...
 		}
 	} else {
-		rv, err = newFaissFloat32Index(fIndex)
-		if err != nil {
-			return nil, nil, nil, fmt.Errorf("faiss float32 index creation error: %v", err)
+		if useGPU {
+			rv, err = newFaissGPUFloat32Index(fIndex)
+			if err != nil {
+				// GPU not available or clone failed, fall back to CPU-only index
+				rv, err = newFaissFloat32Index(fIndex)
```
CascadingRadium: This leaks the internal info out here.

The ideal behaviour would be:

- You always return the `newFaissGPUFloat32Index` struct when `useGPU` flag is set.
- If there is error in the cloning, the `newFaissGPUFloat32Index` will have `gpuIdx` as nil, but keeps the `cpuIdx`.
- The caller should not worry about the internal schematics of the object.
- The object should behave like so:
    - If `gpuIdx` is nil: direct every API call to `cpuIdx`
    - if `gpuIdx` is non-nil: direct specific API calls to `gpuIdx`

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,162 @@
...
+	gpuIdx *faiss.GPUIndexImpl
+}
+
+func newFaissGPUFloat32Index(cpuIdx *faiss.IndexImpl) (faissIndex, error) {
+	if cpuIdx == nil {
+		return nil, ErrNilIndex
+	}
+	gpuIdx, err := faiss.CloneToGPU(cpuIdx)
+	if err != nil {
+		return nil, err
```
CascadingRadium: ditto: the error should be caught here and `faissGPUFloat32Index` should have `nil` as `gpuIdx`

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,162 @@
...
+		gpuIdx: gpuIdx,
+	}, nil
+}
+
+func (f *faissGPUFloat32Index) add(vecs *vectorSet) error {
+	err := f.gpuIdx.Add(vecs.floatData)
+	if err != nil {
+		return err
+	}
+	return f.syncGPUToCPU()
```
CascadingRadium: The `add` and `train` APIs are called on the GPU index, and we end up calling the `sync` method twice. Can we perhaps reduce this to one time? While keeping the struct idempotent also.

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,176 @@
...
+	cpuIdx *faiss.IndexImpl
+	gpuIdx *faiss.GPUIndexImpl
+}
+
+// if cloning to GPU fails, we still allow searches to happen on CPU only
+func newFaissGPUFloat32Index(cpuIdx *faiss.IndexImpl) (faissIndex, error) {
+	if cpuIdx == nil {
+		return nil, ErrNilIndex
+	}
+	// TODO: log GPU clone errors in some way?
```
CascadingRadium: remove this TODO as we are merging this code in next week. As per the spec, no logging will be done here to avoid log spam. Maybe we should change the API contract to just return gpuIdx?  
- If we get an object out of the GPU, we succeeded.
- If we do not, we failed. 

The error here for why the clone failed does not matter here anyway. Can you give your opinion on this? You can change the API signature in go-faiss yourself.

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,176 @@
...
+func (f *faissGPUFloat32Index) setNProbe(nprobe int32) {
+	f.cpuIdx.SetNProbe(nprobe)
+}
+
+func (f *faissGPUFloat32Index) train(trainingData *vectorSet) error {
+	if f.gpuIdx == nil {
+		return f.cpuIdx.Train(trainingData.floatData)
+	}
+
+	if err := f.gpuIdx.Train(trainingData.floatData); err != nil {
```
CascadingRadium:
```go 
return f.gpuIdx.Train(trainingData.floatData)
```

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,176 @@
...
+		return err
+	}
+	return nil
+}
+
+// syncGPUToCPU clones the current GPU index state back to the CPU index,
+// replacing the old CPU index.
+func (f *faissGPUFloat32Index) syncGPUToCPU() error {
```
CascadingRadium, line 259: We use the `syncGPUToCPU` when we finish `add` operation on the GPU index. As per the spec:
- Any error involved in cloning a CPU index to the GPU will be ignored with a silent fallback to usage of CPU index, which we already do.
- Errors involved in cloning a GPU index to CPU must also be handled properly. By returning `err` in L169, we boil the error all the way up to the zapx layer which can cause an async panic. 

Can you suggest a better way to handle the **CloneToCPU** error?

section_faiss_vector_index.go
```diff
@@ -364,7 +365,7 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(sbs []*SegmentBase,
 	freeReconstructedIndexes(reconsIndexes)
 	// create the faiss index to hold the merged data, and add the
 	// reconstructed vectors into it.
-	config := newFaissIndexConfig(faissFP32Index, indexOptimizedFor, dims, metric, nvecs)
+	config := newFaissIndexConfig(faissFP32Index, indexOptimizedFor, dims, metric, nvecs, useGPU)
```
CascadingRadium, line 368: When we have a BIVF optimized index, the float index created here is the backing index. If the user has enabled GPU flag for a BIVF optimized index, then this step essentially uses GPU for the `train` and `add` steps of either a `flat` or `SQ8` index. Although we do train the `SQ8` index, I am not too sure if the ROI in using the GPU for these indexes is a good idea.

section_faiss_vector_index.go
```diff
@@ -572,7 +584,7 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *CountHashWriter) error {
 		}
 		// create the faiss float32 index for the vectors associated with this field and get the
 		// serialized index bytes to be written out to the segment.
-		config := newFaissIndexConfig(faissFP32Index, content.optimizedFor, content.dimension, metric, nvecs)
+		config := newFaissIndexConfig(faissFP32Index, content.optimizedFor, content.dimension, metric, nvecs, content.useGPU)
```
CascadingRadium, line 582: ditto

CascadingRadium (changes requested):

section_faiss_vector_index.go
CascadingRadium: Just looked at the code, the plain SQ8 index does not support GPU cloning.

CascadingRadium:
hey @capemox, please check the new `batchSearch` API in the request batcher module.

- https://github.com/blevesearch/zapx/pull/384

Please add the batcher as an _optional_ object in your struct, with the struct implementing the `batchSearch` API. That way when we call `SearchWithoutIDs` we essentially forward the request to the batcher module via its `search` API. 

When it has finished batching and needs to execute a batch, it will call the `batchSearch` of your GPU struct, which will execute the search on the GPU index. Take special care to NOT involve the batcher if the clone to GPU fails and if the GPU index is unavailable, in which case we do the CPU fallback.

Resolve the comments and fix the merge conflicts ASAP, I am working on resolving merge conflicts on the base `pre_gpu` branch. 

Thanks 
cc @abhinavdangeti

CascadingRadium:
Hey @capemox, i think youre right, I was rebasing over fastmerge on local and should have pushed it later on. Have reverted.

faiss_vector_index_bivf.go
```diff
@@ -251,7 +251,11 @@ func (b *faissBinaryIndex) setNProbe(nprobe int32) {
...
-	return b.binary.Train(trainingData.binaryData)
-}
+func (b *faissBinaryIndex) trainAndAdd(trainingData *vectorSet, vecsToAdd *vectorSet) error {
+	err := b.binary.Train(trainingData.binaryData)
+	if err != nil {
+		return err
+	}
+
+	return b.add(vecsToAdd)
+}
```
CascadingRadium: please ensure to add the new line after the end of every file, do a `go fmt ./...`

faiss_vector_request_batcher.go
```diff
@@ -163,6 +163,8 @@ type coalesceQueue struct {
 	enqueueCh chan *batchRequest
 	// channel for signaling the batcher to stop processing requests and shut down.
 	stopCh chan struct{}
+	// closed when the monitor goroutine has fully exited, allowing stop() to block until done.
```
CascadingRadium, line 170: This was a nice catch @capemox, without the channel, there was a chance of a crash. Please raise this as a separate commit in the `batcher` branch as well. This is to facilitate easier code review process.

section_faiss_vector_index.go
```diff
@@ -440,26 +453,18 @@ func makeFaissIndex(vecs *vectorSet, config *faissIndexConfig) ([]byte, error) {
...
-		err = ivfIndex.train(vecs)
+	} else if sqIndex := index.castSQ(); sqIndex != nil {
+		err = sqIndex.trainAndAdd(vecs, vecs)
 		if err != nil {
 			return nil, err
 		}
-	} else if sqIndex := index.castSQ(); sqIndex != nil {
-		err = sqIndex.train(vecs)
+	} else {
+		err = index.add(vecs)
```
CascadingRadium, line 464: retain the earlier comment 

```
 add the vectors to the index using sequential vector IDs starting
	// from 0 to N-1
```

section_faiss_vector_index.go
```diff
@@ -546,6 +551,17 @@ func determineFloat32IndexToUse(nvecs, nlist int, optimizationType string) strin
 	}
 }
 
+// check if GPU is applicable for the index type and config
+// the only applicable types are: IVF and IVF-SQ8
+// we skip binary, flat and SQ8 for now
+func isGPUApplicable(idx *faiss.IndexImpl) bool {
```
CascadingRadium: i do not think this helper is useful, as its essentially a `IsIVFIndex()` check, and is highly tied to the float32 index type. Rather just have this in L837.

```go
		if cfg.useGPU && idx.IsIVFIndex() {
			return newFaissGPUFloat32Index(idx)
		}
```
Add appropriate commentary there as well

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,282 @@
...
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.gpuIdx = gpuIdx
+	f.batcher = newRequestBatcher(f)
+}
+
+// attempt to add the vectors to the GPU index. If it fails,
+// fallback to the CPU index
+func (f *faissGPUFloat32Index) add(vecs *vectorSet) error {
+	<-f.doneCh
```
CascadingRadium: add a helper like so:

```go
func (f *faissGPUFloat32Index) waitGPU() {
	<-f.doneCh
}
```
every API of the struct can choose to wait or not, instead of the `<-f.doneCh` check everywhere

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,282 @@
...
+// ---------------------------------
+// faissGPUFloat32Index wraps a CPU float32 index alongside a GPU index.
+// The GPU is used for unfiltered searches (no selector), while all
+// other operations (filtered searches, IVF cluster searches, SQ/IVF
+// operations, serialization, etc.) are delegated to the CPU index.
+type faissGPUFloat32Index struct {
+	cpuIdx *faiss.IndexImpl
+
+	// doneCh is closed when initGPU completes. All GPU-operating functions
+	// except searchWithoutIDs must receive from doneCh before proceeding.
```
CascadingRadium: don't add API specific `searchWithoutIDs` comments to the struct fields here. Make sure the comments are agnostic.

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,282 @@
...
+	defer f.mu.Unlock()
+	f.gpuIdx = gpuIdx
+	f.batcher = newRequestBatcher(f)
+}
+
+// attempt to add the vectors to the GPU index. If it fails,
+// fallback to the CPU index
+func (f *faissGPUFloat32Index) add(vecs *vectorSet) error {
+	<-f.doneCh
+	if f.gpuIdx == nil {
```
CascadingRadium, line 85: There is an issue here. As per the struct, the `mu` mutex is supposed to guard access to the gpuIdx. 

If `teardownGPU` and `add` APIs are called concurrently for this struct, we correctly acquire the write lock in `teardownGPU`, but we access f.gpuIdx without a mutex here.

```go
f.mu.RLock()
gpuIdx := f.gpuIdx
f.mu.RUnlock()

if gpuIdx == nil {
// logic
```

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,282 @@
...
+func (f *faissGPUFloat32Index) close() {
+	<-f.doneCh
+	f.teardownGPU()
+	f.cpuIdx.Close()
+}
+
+// teardownGPU stops the batcher first (while gpuIdx is still live so that
+// the final flush can complete on the GPU), then nils and closes the GPU index.
+func (f *faissGPUFloat32Index) teardownGPU() {
+	<-f.doneCh
```
CascadingRadium: I think we always call `teardownGPU()` API after `waitGPU()` is already called in an enclosing API(like `add`, `close`). Although waiting here as well doesn't do us any harm, i think its better to not wait here at all.

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	return f.cpuIdx.Add(vecsToAdd.floatData)
+}
+
+// syncGPUToCPU clones the current GPU index state back to the CPU index,
+// replacing the old CPU index. The write lock is held for the entire duration
+// to block concurrent searchWithoutIDs calls and to protect the cpuIdx swap.
+func (f *faissGPUFloat32Index) syncGPUToCPU() error {
+	f.mu.Lock()
```
CascadingRadium: Do we need to access a write lock here? From what i can see, the `cpuIdx` is not protected by the `mu` mutex right? `mu` only protects `gpuIdx`. I think we can do fine with just a read lock here instead.

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (f *faissGPUFloat32Index) metricType() int {
+	return f.cpuIdx.MetricType()
+}
+
+func (f *faissGPUFloat32Index) searchWithoutIDs(qVector *vectorSet, k int64, selector faiss.Selector, params json.RawMessage) ([]float32, []int64, error) {
+	if selector == nil && len(params) == 0 {
+		f.mu.RLock()
+		if f.batcher != nil && f.gpuIdx != nil {
+			distances, ids, err := f.batcher.search(qVector, k)
```
CascadingRadium, line 139: better to defer here right?

```go
defer f.mu.RUnlock()
return f.batcher.search(qVector, k)
```

CascadingRadium (commented):
first pass

faiss_vector_index_gpu_float32.go
capemox: I actually didn't want to acquire a write lock in `teardownGPU` either, I'll fix that. The purpose of the lock is only to control access to the gpu pointer while the gpu init runs, and so this involves only the `initGPU` and `searchWithoutIDs` methods. Since all other methods wait for the move to finish, we don't need the lock.

faiss_vector_index_gpu_float32.go
capemox: Why should the RLock be held until the function returns? It only needs to hold it until the gpu related operations are finished, at which point it's released.

faiss_vector_index_gpu_float32.go
capemox: Done

faiss_vector_index_gpu_float32.go
CascadingRadium: fair point.

faiss_vector_index_gpu_float32.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func (f *faissGPUFloat32Index) setDirectMap(directMapType int) error {
+	return f.cpuIdx.SetDirectMap(directMapType)
+}
+
+func (f *faissGPUFloat32Index) setNProbe(nprobe int32) {
+	f.cpuIdx.SetNProbe(nprobe)
+	f.waitGPU()
+	if f.gpuIdx != nil {
```
CascadingRadium: I dont think GPU index supports this API. We should be setting nprobe to only the cpu index, since this is in the indexing path and we serialize the cpu index only anyway.

CascadingRadium (commented):
Please pull in minor fixes from the `batcher` branch as well.

CascadingRadium (approved):

PR #388, opened 2026-03-31
MB-59670, MB-62182: [v17] Refactor `faissIndex` into a polymorphic interface

CascadingRadium:
- Replaced the monolithic `faissIndex` struct with a `faissIndex` interface, with `faissFloat32Index` and `faissBinaryIndex` as concrete implementations.
- Eliminated pervasive if `bIndex != nil` branching across all search paths.
- Introduced `vectorSet` type to wrap float/binary vector data, with lazy binarization.
- Added a unified `faissIndexFactory` to centralize index construction.
- Requires https://github.com/blevesearch/go-faiss/pull/51.
- Fix few bugs in the `fastmerge` path.


faiss_vector_cache.go
```diff
@@ -139,34 +139,40 @@ func (vc *vectorIndexCache) createAndCacheLOCKED(fieldID uint16, mem []byte,
 		return nil, nil, nil, fmt.Errorf("could not read faiss index size")
 	}
 	pos += n
-	index = &faissIndex{}
-
+	var rv faissIndex
```
Likith101: nit: Retain the name 'index' here so it is more readable

faiss_vector_index.go
```diff
@@ -0,0 +1,111 @@
+//  Copyright (c) 2024 Couchbase, Inc.
```
Likith101: 2026

section_faiss_vector_index.go
```diff
@@ -1066,5 +989,73 @@ func (v *vectorIndexOpaque) Set(key string, val interface{}) {
...
+	f.nlist = nlist
+}
+
+// Factory function to create a faissIndex for the given index config.
+func faissIndexFactory(cfg *faissIndexConfig) (faissIndex, error) {
+	switch cfg.indexType {
+	case faissFP32Index:
+		// check if we have an nlist set, if so we use that directly,
+		// otherwise we determine it based on the number of vectors.
+		var nlist int
```
Likith101: Set cfg.nlist during creation and let setNList() overwrite the value so you can skip these statements

faiss_vector_index.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/blevesearch/go-faiss"
+)
+
+var (
+	ErrNilIndex     error = errors.New("faiss index cannot be nil")
```
abhinavdangeti: Refactor message to what the situation is rather what it can't be - so `faiss index is nil`.

Also, no need to export I suppose?

faiss_vector_index.go
```diff
@@ -0,0 +1,111 @@
...
+	ErrInvalidIndex error = errors.New("input faiss index is invalid or incompatible for the intended operation")
```
abhinavdangeti: Rather than `ErrInvalidIndex` - would recommend `errNotSupported` (based on usage) and the message accordingly.

abhinavdangeti (commented):
Looks good to me, minor comments around error naming.

abhinavdangeti (dismissed):
👍🏼  requires: https://github.com/blevesearch/go-faiss/pull/51 to be merged first.

abhinavdangeti:
@CascadingRadium Update go-faiss to `v1.0.30` - I suppose we're ok to not do this kind of a thing for `v16`?

abhinavdangeti (dismissed):

capemox (dismissed):

faiss_vector_index_bivf.go
```diff
@@ -0,0 +1,276 @@
...
+// ---------------------------------
+// Faiss Binary IVF Index
+// ---------------------------------
+type faissBinaryIndex struct {
+	backing *faiss.IndexImpl
+	binary  *faiss.BinaryIndexImpl
+}
+
+func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl) (index faissIndex, err error) {
+	// the binary index cannot be nil, but the backing index can be nil, depending on the use case.
```
Likith101: Both the indexes always need to be present. There is no usecase where either is nil

Likith101, line 35: According the to the design of BQ. There will never be a case where backing is nil. However, I see that within the code, during index creation you have a case where you do end up sending a nil backing index.

However, I do not see why you need to accommodate this case everywhere with checks, especially within the search paths.

I would suggest documenting that path throughly and removing the nil checks everywhere. Or maybe keeping the nil checks but throwing an error in those cases. Like, in metric type, if it fails, we should be throwing an error because the binary index metric type is hamming distance which we do not track with a variable anywhere.

Likith101 (changes requested):

faiss_vector_index_bivf.go
CascadingRadium: done

Likith101 (approved):

capemox (approved):

abhinavdangeti:
`v17.0.7`

PR #389, opened 2026-03-31
MB-66395: [v17] Add Request Batcher Module (for GPU indexes)

CascadingRadium:
- Add latency-aware batching for vector search requests using a coalescing queue 
  to combine compatible requests and execute them together to reduce number of 
  Faiss calls.
- Ensure adaptive batching by using Nagle's algorithm for maximum efficiency.


faiss_vector_request_batcher.go
```diff
@@ -188,9 +191,13 @@ func (q *coalesceQueue) stop() {
 	q.stopOnce.Do(func() {
 		close(q.stopCh)
 	})
+
+	// wait for the monitor to stop running
+	<-q.doneCh
```
CascadingRadium: hey @capemox, shouldn't this line (L196) be within the `stopOnce` routine? At L192.5?

capemox: If the wait was done inside, and two concurrent stop() calls were made, only the call that enters the `Once.Do` would wait while the other would return immediately, which can cause issues.

This will make sure both routines wait until the batcher is closed.

CascadingRadium: ah ok thanks!

CascadingRadium:
exploring a self-tuning algorithm for higher QPS while batching. Converting to draft.

faiss_vector_request_batcher.go
```diff
@@ -0,0 +1,292 @@
...
+)
+
+// The requestBatcher is responsible for batching search requests to a Faiss index.
+// It will accumulate incoming search requests and execute them in batches to improve performance.
+// The batcher will use the provided Faiss index to perform the searches, and it
+// will manage the batching logic, including timing and concurrency control.
+type requestBatcher struct {
+	// the coalesce queue that manages the batching of incoming search requests.
+	cq *coalesceQueue
+}
```
abhinavdangeti, line 36: Let's add some unit testing and/or benchmarks for how many requests we can execute in a given window of time (assuming a constant runtime for each batch).

abhinavdangeti (changes requested):

faiss_vector_index.go
```diff
@@ -103,9 +103,18 @@ type faissIndexIVF interface {
...
+// Interface for SQ-specific operations on Faiss vector indices.
 type faissIndexSQ interface {
 	// trains the SQ index on the provided training data. This step is necessary to
 	// perform quantization of the vector space, which enables efficient storage and
 	// search of high-dimensional vectors.
 	train(trainingData *vectorSet) error
 }
+
+// Interface for batched search operations on Faiss vector indices.
+type faissIndexBatch interface {
```
Likith101: nit: faissQueryBatch

Likith101 (changes requested):

faiss_vector_index.go
CascadingRadium: done

faiss_vector_request_batcher.go
CascadingRadium: done

faiss_vector_request_batcher.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	return append(queue, req)
+}
+
+// executeBatch runs all coalesced requests against the Faiss index and delivers results.
+func (q *coalesceQueue) executeBatch(batch []*batchRequest) {
+	for _, req := range batch {
+		distances, ids, err := q.idx.batchSearch(req.qVector, req.k)
+		req.sendResponse(distances, ids, err)
+	}
```
capemox, line 289: @CascadingRadium do you think these can be parallelized across different goroutines?

It might violate the single thread access of the gpu resources, but I wanted to check with you.

CascadingRadium: Hey, no you cannot do that, the GPU indices must be queried using a single thread. Running parallel batches will be equivalent to running multi-threaded queries.

capemox (approved):

abhinavdangeti (approved):

Likith101 (approved):

faiss_vector_request_batcher.go
```diff
@@ -0,0 +1,292 @@
...
+func (m *batchManager) putBatch(batch []*batchRequest) {
+	clear(batch)
+	m.batchPool.Put(batch[:0])
+}
+
+// --------------------------------------------------
+// coalesceQueue
+// --------------------------------------------------
+// Implements Nagle's algorithm for coalescing search requests:
+//   - The coalesce goroutine continuously receives and coalesces incoming requests.
```
Thejas-bhat, line 178: nit: *filler goroutine?

PR #390, opened 2026-04-01
bug: use `index.TrainedIndexCallbackFn`

Thejas-bhat:
- keep the type of func signatures same across bleve and zap

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -64,12 +64,6 @@ const (
 	faissBIVFIndex
 )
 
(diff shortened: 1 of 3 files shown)


section_faiss_vector_index.go
abhinavdangeti: Drop this now shouldn't we?

abhinavdangeti:
Also go.mod update to use bleve_index_api@v1.3.9.

abhinavdangeti (approved):

abhinavdangeti:
`v17.0.8`

PR #393, opened 2026-04-09
MB-71216: implement fast merge for binary index class

Thejas-bhat:
- implement the `mergeFrom` and `setQuantizers` APIs for binary class indexes.
- this PR includes fast merge support for
    - memory-efficient and rabitq optimized indexes
    - bivf indexes having sq8 as the backing index type
- the fast merge path is hit only if both the trainedIndex and the finalMergedIndex are eligible for fast merge, this check ensures that we don't error out and lose data as a fallback mechanism.
- in case of updates/deletes on the index, we avoid re-training the index since that at a later point in time we might end up merging indexes with different centroid layout - that'll cause a panic.

### DIFF
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -94,6 +94,9 @@ type faissIndexIVF interface {
 	// IVF index that is trained on the same data and used to assign vectors
 	// to clusters in the IVF index.
 	setQuantizers(trainedIndex faissIndexIVF) error
+	// quantization return the type of quantization used which can help determing if
+	// the fast merge path is applicable or not using the trained index
+	quantization() string
 	// merged another faiss index into the current IVF index,
 	// with an offset to adjust vector IDs from the other index.
 	mergeFrom(other faissIndex, offset int64) error
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -41,8 +41,8 @@ type faissIndex interface {
 	metricType() int
 	// ntotal returns the total number of vectors currently stored in the index.
 	ntotal() int64
-	// based on the optimization type and the index details we decide and return a
-	// boolean indicating whether the index is eligible for fast merge.
+	// based on the optimization type and the index details we decide and return
+	// whether the participating child index is eligible to be fast merged or not
 	isMergeable(optimizedFor string) bool
 	// reconstructBatch reconstructs the original vectors for the given vector IDs in the index.
 	reconstructBatch(vecIDs []int64, prealloc []float32) ([]float32, error)
@@ -94,8 +94,8 @@ type faissIndexIVF interface {
 	// IVF index that is trained on the same data and used to assign vectors
 	// to clusters in the IVF index.
 	setQuantizers(trainedIndex faissIndexIVF) error
-	// quantization return the type of quantization used which can help determing if
-	// the fast merge path is applicable or not using the trained index
+	// quantization returns the type of quantization of the trained IVF index based
+	// on which we can determine if its viable to be used for fast merge or not
 	quantization() string
 	// merged another faiss index into the current IVF index,
 	// with an offset to adjust vector IDs from the other index.
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -42,8 +42,6 @@ type faissIndex interface {
 	metricType() int
 	// ntotal returns the total number of vectors currently stored in the index.
 	ntotal() int64
-	// returns whether index is eligible for fast merge
-	isMergeable() bool
 	// reconstructBatch reconstructs the original vectors for the given vector IDs in the index.
 	reconstructBatch(vecIDs []int64, prealloc []float32) ([]float32, error)
 	// performs a search on the index using the provided query vector and and retrieves the top K nearest neighbors.
@@ -95,6 +93,8 @@ type faissIndexIVF interface {
 	// IVF index that is trained on the same data and used to assign vectors
 	// to clusters in the IVF index.
 	setQuantizers(trainedIndex faissIndexIVF) error
+	// returns whether the participating index is eligible for fast merge
+	isMergeable() bool
 	// merged another faiss index into the current IVF index,
 	// with an offset to adjust vector IDs from the other index.
 	mergeFrom(other faissIndex, offset int64) error
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -41,6 +41,9 @@ type faissIndex interface {
 	metricType() int
 	// ntotal returns the total number of vectors currently stored in the index.
 	ntotal() int64
+	// based on the optimization type and the index details we decide and return a
+	// boolean indicating whether the index is eligible for fast merge.
+	isMergeable(optimizedFor string) bool
 	// reconstructBatch reconstructs the original vectors for the given vector IDs in the index.
 	reconstructBatch(vecIDs []int64, prealloc []float32) ([]float32, error)
 	// performs a search on the index using the provided query vector and and retrieves the top K nearest neighbors.
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -25,6 +25,7 @@ import (
 )
 
 var (
+	errNilConfig    error = errors.New("faiss index config is nil")
 	errNilIndex     error = errors.New("faiss index is nil")
 	errNotSupported error = errors.New("operation not supported")
 )
@@ -41,9 +42,8 @@ type faissIndex interface {
 	metricType() int
 	// ntotal returns the total number of vectors currently stored in the index.
 	ntotal() int64
-	// based on the optimization type and the index details we decide and return
-	// whether the participating child index is eligible to be fast merged or not
-	isMergeable(optimizedFor string) bool
+	// returns whether index is eligible for fast merge
+	isMergeable() bool
 	// reconstructBatch reconstructs the original vectors for the given vector IDs in the index.
 	reconstructBatch(vecIDs []int64, prealloc []float32) ([]float32, error)
 	// performs a search on the index using the provided query vector and and retrieves the top K nearest neighbors.
@@ -64,6 +64,7 @@ type faissIndex interface {
 
 // Interface for IVF-specific operations on Faiss vector indices.
 type faissIndexIVF interface {
+	faissIndex
 	// returns the count of the selected vector IDs in each
 	// cluster of the IVF index, based on the provided selector.
 	clusterVectorCounts(sel faiss.Selector, nlist int) ([]int64, error)
@@ -94,9 +95,6 @@ type faissIndexIVF interface {
 	// IVF index that is trained on the same data and used to assign vectors
 	// to clusters in the IVF index.
 	setQuantizers(trainedIndex faissIndexIVF) error
-	// quantization returns the type of quantization of the trained IVF index based
-	// on which we can determine if its viable to be used for fast merge or not
-	quantization() string
 	// merged another faiss index into the current IVF index,
 	// with an offset to adjust vector IDs from the other index.
 	mergeFrom(other faissIndex, offset int64) error
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -90,7 +90,7 @@ type faissIndexIVF interface {
 	// sets the quantizers for the IVF index. The quantizer is a separate
 	// IVF index that is trained on the same data and used to assign vectors
 	// to clusters in the IVF index.
-	setQuantizers(centroidIndex faissIndexIVF) error
+	setQuantizers(trainedIndex faissIndexIVF) error
 	// merged another faiss index into the current IVF index,
 	// with an offset to adjust vector IDs from the other index.
 	mergeFrom(other faissIndex, offset int64) error
--- a/faiss_vector_index_bivf.go
+++ b/faiss_vector_index_bivf.go
@@ -49,15 +49,15 @@ func newFaissBinaryIndexWithConfig(binary *faiss.BinaryIndexImpl, backing *faiss
 	if binary == nil || backing == nil {
 		return nil, errNilIndex
 	}
+	if cfg == nil {
+		return nil, errNilConfig
+	}
+
 	return &faissBinaryIndex{
 		cfg:     cfg,
 		backing: backing,
 		binary:  binary,
 	}, nil
-}
-
-func (b *faissBinaryIndex) quantization() string {
-	return quantizationType(b.cfg.optimizationType, b.binary.Ntotal())
 }
 
 func (b *faissBinaryIndex) add(vecs *vectorSet) error {
@@ -286,13 +286,21 @@ func (b *faissBinaryIndex) setQuantizers(trainedIndex faissIndexIVF) error {
 	return errNotSupported
 }
 
-func (b *faissBinaryIndex) isMergeable(optimizedFor string) bool {
-	// note: currently the only the bivf-sq8 type indexes are eligible for fast merge
-	return b.binary.IsIVFIndex() && optimizedFor == index.IndexBIVFWithBackingSQ8
+func (b *faissBinaryIndex) isMergeable() bool {
+	if b.cfg != nil {
+		switch b.cfg.optimizationType {
+		case index.IndexBIVFWithBackingFlat, index.IndexBIVFWithBackingSQ8:
+			return b.backing.Ntotal() > 1000
+		}
+	}
+	return false
 }
 
 func (b *faissBinaryIndex) mergeFrom(other faissIndex, offset int64) error {
 	if idx, ok := other.(*faissBinaryIndex); ok {
+		if !idx.isMergeable() {
+			return errNotSupported
+		}
 		// merge the binary and the backing index, both flat and SQ8 indexes support
 		// merge_from API underneath the hood. the add_id is kept to 0 since we will
 		// be merging the largest set of indexes which will be sequential in the list
--- a/faiss_vector_index_bivf.go
+++ b/faiss_vector_index_bivf.go
@@ -29,6 +29,7 @@ import (
 // Faiss Binary IVF Index
 // ---------------------------------
 type faissBinaryIndex struct {
+	cfg     *faissIndexConfig
 	backing *faiss.IndexImpl
 	binary  *faiss.BinaryIndexImpl
 }
@@ -42,6 +43,21 @@ func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl
 		backing: backing,
 		binary:  binary,
 	}, nil
+}
+
+func newFaissBinaryIndexWithConfig(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl, cfg *faissIndexConfig) (index faissIndex, err error) {
+	if binary == nil || backing == nil {
+		return nil, errNilIndex
+	}
+	return &faissBinaryIndex{
+		cfg:     cfg,
+		backing: backing,
+		binary:  binary,
+	}, nil
+}
+
+func (b *faissBinaryIndex) quantization() string {
+	return quantizationType(b.cfg.optimizationType, b.binary.Ntotal())
 }
 
 func (b *faissBinaryIndex) add(vecs *vectorSet) error {
--- a/faiss_vector_index_bivf.go
+++ b/faiss_vector_index_bivf.go
@@ -289,8 +289,11 @@ func (b *faissBinaryIndex) setQuantizers(trainedIndex faissIndexIVF) error {
 func (b *faissBinaryIndex) isMergeable() bool {
 	if b.cfg != nil {
 		switch b.cfg.optimizationType {
-		case index.IndexBIVFWithBackingFlat, index.IndexBIVFWithBackingSQ8:
-			return b.backing.Ntotal() > 1000
+		case index.IndexBIVFWithBackingFlat:
+			// the flat backing index currently doesn't support merge_from
+			return false
+		case index.IndexBIVFWithBackingSQ8:
+			return b.backing.Ntotal() > ivfThreshold
 		}
 	}
 	return false
--- a/faiss_vector_index_bivf.go
+++ b/faiss_vector_index_bivf.go
@@ -21,6 +21,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 
+	index "github.com/blevesearch/bleve_index_api"
 	faiss "github.com/blevesearch/go-faiss"
 )
 
@@ -267,6 +268,11 @@ func (b *faissBinaryIndex) setQuantizers(trainedIndex faissIndexIVF) error {
 		return nil
 	}
 	return errNotSupported
+}
+
+func (b *faissBinaryIndex) isMergeable(optimizedFor string) bool {
+	// note: currently the only the bivf-sq8 type indexes are eligible for fast merge
+	return b.binary.IsIVFIndex() && optimizedFor == index.IndexBIVFWithBackingSQ8
 }
 
 func (b *faissBinaryIndex) mergeFrom(other faissIndex, offset int64) error {
--- a/faiss_vector_index_bivf.go
+++ b/faiss_vector_index_bivf.go
@@ -254,7 +254,7 @@ func (b *faissBinaryIndex) setQuantizers(trainedIndex faissIndexIVF) error {
 	if idx, ok := trainedIndex.(*faissBinaryIndex); ok {
 		// set quantizers for the binary and the backing index if its an SQ8 index
 		var err error
-		if idx.backing != nil && idx.backing.IsSQIndex() {
+		if idx.backing.IsSQIndex() {
 			err = b.backing.SetQuantizers(idx.backing)
 			if err != nil {
 				return err
--- a/faiss_vector_index_bivf.go
+++ b/faiss_vector_index_bivf.go
@@ -251,11 +251,40 @@ func (b *faissBinaryIndex) trainAndAdd(trainingData *vectorSet, vecsToAdd *vecto
 }
 
 func (b *faissBinaryIndex) setQuantizers(trainedIndex faissIndexIVF) error {
-	// not supported for binary indexes, returns error
+	if idx, ok := trainedIndex.(*faissBinaryIndex); ok {
+		// set quantizers for the binary and the backing index if its an SQ8 index
+		var err error
+		if idx.backing != nil && idx.backing.IsSQIndex() {
+			err = b.backing.SetQuantizers(idx.backing)
+			if err != nil {
+				return err
+			}
+		}
+		err = b.binary.SetQuantizers(idx.binary)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
 	return errNotSupported
 }
 
 func (b *faissBinaryIndex) mergeFrom(other faissIndex, offset int64) error {
-	// merging is not supported for binary indexes, return error
+	if idx, ok := other.(*faissBinaryIndex); ok {
+		// merge the binary and the backing index, both flat and SQ8 indexes support
+		// merge_from API underneath the hood. the add_id is kept to 0 since we will
+		// be merging the largest set of indexes which will be sequential in the list
+		// of segments being merged, so there won't be any ID conflicts.
+		err := b.backing.MergeFrom(idx.backing, 0)
+		if err != nil {
+			return err
+		}
+		err = b.binary.MergeFrom(idx.binary, offset)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
 	return errNotSupported
 }
--- a/faiss_vector_index_float32.go
+++ b/faiss_vector_index_float32.go
@@ -29,6 +29,7 @@ import (
 // Faiss Float32 Index
 // ---------------------------------
 type faissFloat32Index struct {
+	cfg *faissIndexConfig
 	idx *faiss.IndexImpl
 }
 
@@ -39,6 +40,20 @@ func newFaissFloat32Index(idx *faiss.IndexImpl) (index faissIndex, err error) {
 	return &faissFloat32Index{
 		idx: idx,
 	}, nil
+}
+
+func newFaissFloat32IndexWithConfig(idx *faiss.IndexImpl, cfg *faissIndexConfig) (index faissIndex, err error) {
+	if idx == nil {
+		return nil, errNilIndex
+	}
+	return &faissFloat32Index{
+		idx: idx,
+		cfg: cfg,
+	}, nil
+}
+
+func (f *faissFloat32Index) quantization() string {
+	return quantizationType(f.cfg.optimizationType, f.idx.Ntotal())
 }
 
 func (f *faissFloat32Index) add(vecs *vectorSet) error {
@@ -158,9 +173,12 @@ func (f *faissFloat32Index) setQuantizers(trainedIndex faissIndexIVF) error {
 }
 
 func (f *faissFloat32Index) isMergeable(optimizedFor string) bool {
-	// only the ones that are of IVF index family and optimized for IVF RabitQ or
-	// are large enough to have SQ8 quantization are eligible for fast merge
-	return f.idx.IsIVFIndex() && (optimizedFor == index.IndexIVFRaBitQ || f.ntotal() >= ivfSq8Threshold)
+	// merge criteria:
+	// - optimization is ivf,rabitq
+	// - optimization is memory efficient
+	// - total vecs > ivfSq8Threshold => SQ8 quantization is used
+	return f.idx.IsIVFIndex() && (optimizedFor == index.IndexIVFRaBitQ ||
+		optimizedFor == index.IndexOptimizedForMemoryEfficient || f.ntotal() >= ivfSq8Threshold)
 }
 
 func (f *faissFloat32Index) mergeFrom(other faissIndex, offset int64) error {
--- a/faiss_vector_index_float32.go
+++ b/faiss_vector_index_float32.go
@@ -21,6 +21,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 
+	index "github.com/blevesearch/bleve_index_api"
 	faiss "github.com/blevesearch/go-faiss"
 )
 
@@ -154,6 +155,12 @@ func (f *faissFloat32Index) setQuantizers(trainedIndex faissIndexIVF) error {
 		return errNotSupported
 	}
 	return f.idx.SetQuantizers(centroidFaissIndex.idx)
+}
+
+func (f *faissFloat32Index) isMergeable(optimizedFor string) bool {
+	// only the ones that are of IVF index family and optimized for IVF RabitQ or
+	// are large enough to have SQ8 quantization are eligible for fast merge
+	return f.idx.IsIVFIndex() && (optimizedFor == index.IndexIVFRaBitQ || f.ntotal() >= ivfSq8Threshold)
 }
 
 func (f *faissFloat32Index) mergeFrom(other faissIndex, offset int64) error {
--- a/faiss_vector_index_float32.go
+++ b/faiss_vector_index_float32.go
@@ -175,17 +175,14 @@ func (f *faissFloat32Index) setQuantizers(trainedIndex faissIndexIVF) error {
 func (f *faissFloat32Index) isMergeable() bool {
 	if f.cfg != nil {
 		switch f.cfg.optimizationType {
-		case index.IndexBIVFWithBackingFlat:
-			// currently if the backing index is flat, we're not going to perform fast merge
-			return false
-		case index.IndexBIVFWithBackingSQ8:
+		case index.IndexBIVFWithBackingFlat, index.IndexBIVFWithBackingSQ8:
 			fallthrough
 		case index.IndexOptimizedForMemoryEfficient:
 			fallthrough
 		case index.IndexIVFRaBitQ:
-			return f.idx.Ntotal() > 1000
+			return f.idx.Ntotal() > ivfThreshold
 		default:
-			return f.idx.Ntotal() > 10000
+			return f.idx.Ntotal() > ivfSq8Threshold
 
 		}
 	}
--- a/faiss_vector_index_float32.go
+++ b/faiss_vector_index_float32.go
@@ -46,14 +46,14 @@ func newFaissFloat32IndexWithConfig(idx *faiss.IndexImpl, cfg *faissIndexConfig)
 	if idx == nil {
 		return nil, errNilIndex
 	}
+	if cfg == nil {
+		return nil, errNilConfig
+	}
+
 	return &faissFloat32Index{
 		idx: idx,
 		cfg: cfg,
 	}, nil
-}
-
-func (f *faissFloat32Index) quantization() string {
-	return quantizationType(f.cfg.optimizationType, f.idx.Ntotal())
 }
 
 func (f *faissFloat32Index) add(vecs *vectorSet) error {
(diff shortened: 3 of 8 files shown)


abhinavdangeti:
After a conversation w/ @Thejas-bhat , he has another patch coming here soon - to address the situation where we would need to merge segments that have deleted items set in the bitmap. This will be applicable to all indexes participating in the fast merge.

- A plan would be to reconstruct vectors and introduce them into a new index that uses the centroid template.
- For RabitQ - we'd be attempting reconstruction as well .. which would result in loss of precision.

In both situations above, data drift will likely occur - which is fine because `vector_index_fast_merge` is meant to work well over static datasets.

faiss_vector_index.go
```diff
@@ -46,9 +46,9 @@ type faissIndex interface {
 	// performs a search on the index using the provided query vector and and retrieves the top K nearest neighbors.
 	// Optional search constraints can be applied using the selector and additional search parameters.
 	search(qVector *vectorSet, k int64, selector faiss.Selector, params json.RawMessage) ([]float32, []int64, error)
-	// serializes the index into a byte slice,
-	// which can be stored or transmitted.
-	serialize() ([]byte, error)
+	// write out the index content into the provide fileWriter using a reusable buffer
+	// returns the bytes written and any error encountered during the write process.
```
CascadingRadium: comment is wrong, we only return the error not bytes written

section_faiss_vector_index.go
```diff
@@ -1061,9 +1039,15 @@ func faissIndexFactory(cfg *faissIndexConfig) (faissIndex, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		description = determineFloat32IndexToUse(cfg.numVecs, cfg.nlist, cfg.optimizationType)
+		backing, err := faiss.IndexFactory(cfg.dimension, description, cfg.metricType)
+		if err != nil {
+			return nil, err
+		}
 		// set no backing index on purpose as we use the factory in the
```
CascadingRadium: comment needs to be removed as we are inserting the backing index here.

section_faiss_vector_index.go
```diff
@@ -145,24 +149,28 @@ func (v *faissVectorIndexSection) Merge(opaque map[int]resetable, segments []*Se
 			if pos == 0 {
 				continue
 			}
+
```
CascadingRadium, line 151: avoid empty lines in the PR

section_faiss_vector_index.go
```diff
@@ -1061,9 +1039,15 @@ func faissIndexFactory(cfg *faissIndexConfig) (faissIndex, error) {
...
+		description = determineFloat32IndexToUse(cfg.numVecs, cfg.nlist, cfg.optimizationType)
+		backing, err := faiss.IndexFactory(cfg.dimension, description, cfg.metricType)
+		if err != nil {
+			return nil, err
+		}
 		// set no backing index on purpose as we use the factory in the
 		// indexing path, where each index is handled independently
-		return newFaissBinaryIndex(idx, nil)
+		return newFaissBinaryIndex(idx, backing)
```
CascadingRadium: rename `idx` to `binary`

faiss_vector_index_bivf.go
CascadingRadium: We no longer allow creation of `faissBinaryIndex` without a backing index following this refactor. 

Add a hard validation here.

```go
	if binary == nil || backing == nil {
         return nil, errNilIndex
	}
```

Remove all  `if b.backing != nil {` checks from this file, as we no longer allow object construction without the backing index. We can follow that invariant.

section_faiss_vector_index.go
```diff
@@ -580,109 +628,79 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(centroidIndex faissIndexI
...
 	// ensure the faiss index is closed after use
 	defer index.close()
+
+	if config.indexType == faissBIVFIndex {
+		vecs.binarize()
+	}
 	// if we are using an IVF index, train and add first, then set the direct map
 	// and nprobe. The order matters for GPU indexes: CloneToCPU (done inside
 	// trainAndAdd) clears the direct map and nprobe, so they must be set after.
+	// binarize the vectors for BIVF indexes
```
CascadingRadium: misplaced comment, should be in L652

section_faiss_vector_index.go
```diff
@@ -580,109 +628,79 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(centroidIndex faissIndexI
...
 		// future merges use direct map type 1 -> array based direct map, since
 		// we have sequential vector IDs starting from 0 to N-1.
 		err = ivfIndex.setDirectMap(1)
 		if err != nil {
-			return nil, err
+			return err
 		}
 		// calculate nprobe using a heuristic.
 		nprobe := calculateNprobe(config.nlist, config.optimizationType)
 		ivfIndex.setNProbe(nprobe)
```
CascadingRadium, line 649: hmm

the old code basically created the float and binary indexes separately so we needed to call `trainAndAdd` for the SQ backing index. 

Now every BIVF index enters the `castIVF()` block (since we are providing the `faissBinaryIndex` here. 

When the backing index is:
- flat => only add
- SQ => train and add

in the `trainAndAdd` of `faissBinaryIndex` I see train being called always. Need to have the object keep this discerning property.

section_faiss_vector_index.go
```diff
@@ -209,16 +220,27 @@ func (v *faissVectorIndexSection) Merge(opaque map[int]resetable, segments []*Se
...
 			pos += int(indexSize)
+
+			// track the start offset and size of the binary index if present
+			if newIndexInfo.indexType == faissBIVFIndex {
+				binSize, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+				pos += n
+				newIndexInfo.binaryIndexStartOffset = pos
+				newIndexInfo.binaryIndexSize = binSize
+				pos += int(binSize)
+			}
```
CascadingRadium: I think we should avoid indexType specific data here. I think we can trim `vecIndexInfo` to only have `startOffset` and `indexType`. The `binary*` and `indexSize` fields are redundant and can be removed. When we are in `mergeAndWriteVectorIndexes` method, we can use the `indexType` and `startOffset` to determine the required details on the fly.

section_faiss_vector_index.go
```diff
@@ -517,6 +546,27 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(centroidIndex faissIndexI
...
+				return err
+			}
+
+			binaryIndex, err := faiss.ReadBinaryIndexFromBuffer(indexBytes, ioFlags)
+			if err != nil {
+				freeReconstructedIndexes(vecIndexes)
+				return err
+			}
+
+			vecIndexes[segI].binaryIndex, err = newFaissBinaryIndex(binaryIndex, faissIndex)
```
CascadingRadium: we end up storing 

```go
vecIndexes[segI].index, err = newFaissFloat32Index(faissIndex)
vecIndexes[segI].binaryIndex, err = newFaissBinaryIndex(binaryIndex, faissIndex)
```

why not have only `index` and store either the fp32 index or the binary index in it based on the index type?

CascadingRadium (changes requested):
first pass

CascadingRadium:
Also hey @Thejas-bhat, i see that this PR has 2 parts.
- Refactor of BIVF code in the reconstruct merge path (with the compulsory backing index in 
  `faissIndexBinary`
- Support of fast merge over the refactored bivf code.

Can you please split the PR into 2 parts, with one focusing on the refactor and the other chained on top of it for supporting fast merge?

It will really help the review process as @Likith101 and I would like to ensure the refactor doesn't cause any regression. Thanks

faiss_vector_index.go
Thejas-bhat: done

faiss_vector_index_bivf.go
Thejas-bhat: done

section_faiss_vector_index.go
Thejas-bhat: i added these line breaks because i feel like the code is too cluttered? it feels cleaner with line breaks which keeps the logical parts of the code together. however, if others dont agree with this i'll just remove this

section_faiss_vector_index.go
Thejas-bhat: yeah makes sense, done

section_faiss_vector_index.go
Thejas-bhat: done

section_faiss_vector_index.go
Thejas-bhat: done

section_faiss_vector_index.go
Thejas-bhat: done

section_faiss_vector_index.go
Thejas-bhat: done

section_faiss_vector_index.go
Thejas-bhat: done

Thejas-bhat:
i've created the separate PR for the refactoring, for sanity check the index builds work fine and i've verified the recall doesnt drop. 
https://github.com/blevesearch/zapx/pull/403. i'll rebase this PR over the refactor one

Thejas-bhat:
> A plan would be to reconstruct vectors and introduce them into a new index that uses the centroid template

i've handled this in the latest patch, thanks for catching this

section_faiss_vector_index.go
```diff
@@ -363,31 +381,30 @@ func (v *vectorIndexOpaque) fastMergeIndexes(centroidIndex faissIndexIVF, dims,
...
 		nvecs += vecCount
 	}
 
-	nprobe, nlist := centroidIndex.ivfParams()
-	// set a custom nlist value in the config for the merged index,
-	//  which is same as the centroid index's nlist.
-	cfg := newFaissIndexConfig(indexType, optimizedFor, dims, metric, nvecs, nlist, false)
-	// create the merged index based on the centroid index's config.
+	// create a faissIndex for merged index using nlist and nprobe from trained index's
+	// config and don't use GPU while creating it
```
abhinavdangeti: Would you expand why we shouldn't use GPU here? Also is this a valid worry - overwriting it's config will prevent this segment from moving to GPU in the future even if user requested for it.

Thejas-bhat: yeah i'll expand the in-line comment - but its mainly because the merge_from API isn't support on the GPU side of the faiss code. this is a check we can probably have on the bleve side? i.e. we can't use both GPU and fast merge config at the same time, similar to how we're supposed to check RabitQ + fast merge.

abhinavdangeti (dismissed):

CascadingRadium:
hey @Thejas-bhat, can you please rebase/merge master. Will give my review post. Thanks.

faiss_vector_index_bivf.go
```diff
@@ -251,11 +251,40 @@ func (b *faissBinaryIndex) trainAndAdd(trainingData *vectorSet, vecsToAdd *vecto
 }
 
 func (b *faissBinaryIndex) setQuantizers(trainedIndex faissIndexIVF) error {
-	// not supported for binary indexes, returns error
+	if idx, ok := trainedIndex.(*faissBinaryIndex); ok {
+		// set quantizers for the binary and the backing index if its an SQ8 index
+		var err error
+		if idx.backing != nil && idx.backing.IsSQIndex() {
```
CascadingRadium: avoid `idx.backing != nil`

faiss_vector_index_bivf.go
```diff
@@ -251,11 +251,40 @@ func (b *faissBinaryIndex) trainAndAdd(trainingData *vectorSet, vecsToAdd *vecto
...
 }
 
 func (b *faissBinaryIndex) mergeFrom(other faissIndex, offset int64) error {
-	// merging is not supported for binary indexes, return error
+	if idx, ok := other.(*faissBinaryIndex); ok {
+		// merge the binary and the backing index, both flat and SQ8 indexes support
+		// merge_from API underneath the hood. the add_id is kept to 0 since we will
+		// be merging the largest set of indexes which will be sequential in the list
+		// of segments being merged, so there won't be any ID conflicts.
+		err := b.backing.MergeFrom(idx.backing, 0)
```
CascadingRadium, line 311: do `Flat` Indexes support merge_from? I had assumed they do but I am not finding the implementation for it. 

I am also not getting why we are passing `offset` to `b.binary` and not `b.backing`.

section_faiss_vector_index.go
```diff
@@ -48,7 +48,7 @@ const (
 	nprobeLatencyOptimization = 2
 	// The threshold for number of vectors beyond which we consider fast merging
 	// using faiss's native merge capabilities, instead of reconstructing and adding
-	// vectors one by one. This is currently only applicable for IVFSQ8 indexes.
+	// vectors one by one
 	ivfSq8Threshold = 10000
```
CascadingRadium, line 55: Shouldn't this be renamed, since we are allowing bivf as well.

faiss_vector_index_bivf.go
Thejas-bhat: > do Flat Indexes support merge_from? 

its a little weird actually - i'll create a separate faiss patch to handle this part. theoretically if you see the IndexFlatCodes codepath you're supposed to be able to merge any indexes that implement IndexFlatCodes but they need to be of the same type. 
what's weird is that when we invoke the faiss factory to build a flat index with metric type inner product, the type of index being built is `IndexFlat` which is as per expectation. 
however, when the index is written how, the header part of the serialized data is marked with type `IndexFlatIP` and when we read the index as part of the merge process we have essentially have the type `IndexFlatIP` that doesn't match the `IndexFlat` type of the final merged index and so the fast merge doesn't actually work based on how i assumed it would. 

so for this patch, let's just support when the backing index is SQ8 else we just fallback to naive reconstruction using the trained template.

Thejas-bhat: the underlying faiss code doesn't support add_id for merging the indexes belonging to flat class of indexes. the ids just increment based on the order in which the indexes having the subset of vectors are merged. so its the responsibility of the caller to order the merge_from operations and we do that appropriately on the zap side.

section_faiss_vector_index.go
Thejas-bhat: i think the naming still holds because this threshold check is applicable only for the ivfsq8 type of indexes, for binary class and rabitq indexes the check is done based on the optimization type and not the vector count

section_faiss_vector_index.go
```diff
@@ -647,6 +667,7 @@ func (v *vectorIndexOpaque) writeFaissIndex(vecs *vectorSet, config *faissIndexC
 	// if we are using an IVF index, train and add first, then set the direct map
 	// and nprobe. The order matters for GPU indexes: CloneToCPU (done inside
 	// trainAndAdd) clears the direct map and nprobe, so they must be set after.
+	// binarize the vectors for BIVF indexes
```
CascadingRadium: redundant comment. It is already there in L663

section_faiss_vector_index.go
```diff
@@ -567,15 +589,16 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(centroidIndex faissIndexI
...
 	// - IVFSQ8 indexes beyond a certain size threshold.
-	// - the centroid index is available.
+	// - the trained index is available.
 	// - RabitQ optimization
-	if centroidIndex != nil &&
-		(indexOptimizedFor == index.IndexIVFRaBitQ ||
-			(nvecs >= ivfSq8Threshold && indexType == faissFP32Index)) {
-		err := v.fastMergeIndexes(centroidIndex, dims, metric, indexType, indexOptimizedFor, vecIndexes, w, closeCh)
+	if trainedIndex != nil &&
+		(indexOptimizedFor == index.IndexIVFRaBitQ || nvecs >= ivfSq8Threshold) {
```
CascadingRadium: I feel we are duplicating the check in L385 as well. If the `indexOptimizedFor` is `index.IndexBIVFWithBackingSQ8` then shouldn't we not enter `fastMergeIndexes` method at all?

I would suggest always calling `fastMergeIndexes` API if `trainedIndex != nil`. The API should perform reconstruct merge if the index type is not supported etc and handle the fallbacks. Perhaps you can create a helper method to perform reconstruct merge which can be reused below and in the `fastMergeIndexes` API.

section_faiss_vector_index.go
```diff
@@ -430,22 +449,25 @@ func (v *vectorIndexOpaque) fastMergeIndexes(centroidIndex faissIndexIVF, dims,
 		// point we can be sure of no updates/deletes -> the candidates will be
 		// in a continuous block from the beginning.
 		if i <= finalMergeCandidate {
-			err = ivfIdx.mergeFrom(floatIdx, faissIndex.ntotal())
+			err = ivfIdx.mergeFrom(childIdx, faissIndex.ntotal())
 			if err != nil {
 				return err
 			}
 		} else {
 			// reconstruction will be done on IVFFlat and Flat indexes
```
CascadingRadium: too specific as we may be doing reconstruct for bflat as well.

section_faiss_vector_index.go
```diff
@@ -363,31 +382,31 @@ func (v *vectorIndexOpaque) fastMergeIndexes(centroidIndex faissIndexIVF, dims,
 		vecCount := len(vecIndexes[i].vecIds)
 		if vecCount == 0 {
 			continue
-		} else if (vecIndexes[i].indexOptimizedFor == index.IndexIVFRaBitQ &&
-			vecIndexes[i].index.castIVF() != nil) ||
-			vecCount >= ivfSq8Threshold {
-			// merging only the large indexes (IVFSQ8 family)
-			// all IVF<same class> indexes are eligible for merging
-			//
+		} else if !drops && v.canMerge(vecIndexes[i], cfg.indexType) {
```
CascadingRadium: I think a better algorithm to use here for `fastMergeIndexes` would be like so:

First we add a castFastMerge API similar to castIVF:

```go

type faissIndex interface {
// ...
	// returns the underlying IVF index if this is an IVF index,
	// nil otherwise.
	castIVF() faissIndexIVF
	// returns the mergeable interface if this index supports fast merge
	// for the given optimization type, nil otherwise.
	castMergeable(optimizedFor string) faissIndexMergeable
}


// faissIndexMergeable is implemented by indexes that support fast merge operations
// via setQuantizers and mergeFrom. Used in the fast merge path.
type faissIndexMergeable interface {
	// adds the given vectors to the index.
	add(vecs *vectorSet) error
	// sets the quantizers for the index. The quantizer is used to assign
	// vectors to clusters.
	setQuantizers(trainedIndex faissIndexMergeable) error
	// mergeFrom merges another faiss index into the current index,
	// with an offset to adjust vector IDs from the other index.
	mergeFrom(other faissIndexMergeable, offset int64) error
}

func (f *faissFloat32Index) castMergeable(optimizedFor string) faissIndexMergeable {
	if !f.idx.IsIVFIndex() {
		return nil
	}
	// FP32 IVF indexes support fast merge for RaBitQ or when above the SQ8 threshold
	if optimizedFor == index.IndexIVFRaBitQ || f.ntotal() >= ivfSq8Threshold {
		return f
	}
	return nil
}

func (b *faissBinaryIndex) castMergeable(optimizedFor string) faissIndexMergeable {
	if !b.binary.IsIVFIndex() {
		return nil
	}
	// BIVF indexes support fast merge only for SQ8 backing indexes
	if optimizedFor == index.IndexBIVFWithBackingSQ8 {
		return b
	}
	return nil
}

func (f *faissGPUFloat32Index) castMergeable(optimizedFor string) faissIndexMergeable {
	// GPU indexes don't support fast merge
	return nil
}
```

new algo:

```go
input - trainedIndex, vecIndexes

  let reconsIndex faissIndex = []
  let mergeIndex faissIndexMergeable = []
  
  for index in vecIndexes:
      if fastMergeIndex := index.castFastMerge()
           mergeIndex.append(fastMergeIndex)
      else 
           reconsIndex.append(index)

  let finalMergedIndex = final mergeable index
  if mergeIndex is not empty:
      set the quantizer using the trained index
      for midx in mergeIndex:
          finalMergedIndex.mergeWith(midx)
  if reconsIndex is not empty:
      reconsAndAddTo(reconsIndex, finalMergedIndex)

  finalMergedIndex.write()
```

We can have `reconsAndAddTo` as a reusable helper for both the fast merge + reconsmerge path.

section_faiss_vector_index.go
CascadingRadium: If we always do fast_merge for bivf (if enabled) and avoid for bflat, why cant we follow this for float32? We follow a similar logic there right? Why do we need an additional check for `ivfSq8Threshold`

section_faiss_vector_index.go
Thejas-bhat: > If the indexOptimizedFor is index.IndexBIVFWithBackingSQ8 then shouldn't we not enter fastMergeIndexes method at all

that's not entirely true because we can still leverage the existing trained index's centroid layout to avoid the re-training part. it'll also help in maintaining the centroid layout consistent in all segments. i'll refactor using your suggestion though.

section_faiss_vector_index.go
Thejas-bhat: the flat and ivfflat indexes aren't fast merged because they're either too small to have the benefit or the quantization doesn't match. like the trained index is something that's supposed to be trained on a big volume that it creates an ivfsq8 trained index. so we can't do a fast merge of ivfflat indexes into the IVFSQ8 index. 

you can suggest any other name that's relevant though

section_faiss_vector_index.go
Thejas-bhat: okay let me refactor the code.

section_faiss_vector_index.go
CascadingRadium: oh sorry i forgot that we have ivf flat also. 
I had meant we do fast merge if IVF else not. But i think with IVF Flat we would need an additional check. resolving

section_faiss_vector_index.go
Thejas-bhat: in case of naive/reconstruct merge path we're accumulating all the vectors from all the indexes and then performing a training followed by add. 

in fast merge path, we strictly avoid a re-training because we might end up with a different centorid layout and cause crashes down the line. also, we do the reconstruct+add for the non-eligible indexes and also the ones that failed the merge_from API call as a fallback mechanism. 

since they're 2 very different type of merging i feel like it's better to leave the reconstruct+add part as is

section_faiss_vector_index.go
Thejas-bhat: done

section_faiss_vector_index.go
Thejas-bhat: done

section_faiss_vector_index.go
Thejas-bhat: i do agree that we can shift the mergeability decision to the `faissIndex` level though, so i've included that part in the latest patch.

section_faiss_vector_index.go
Thejas-bhat: also the latest patch handles the scenario for sq4 quantization (which is applicable for all vector count) and the fast merge path is hit only if the backing quantization types match for both the trained index and the final merged index. 

just having a `trainedIndex != nil` check would cause issues when the trained index is SQ8 and the merged index is of type IVF, Flat since we wouldn't have the count check. The purpose of the count check was to match the quantization type, and in the latest patch i'm handling this in a more sensible manner.

section_faiss_vector_index.go
CascadingRadium: the new algorithm reads a lot better now and the error fallback is a nice touch. Resolving

section_faiss_vector_index.go
```diff
@@ -352,42 +349,38 @@ func calculateNprobe(nlist int, indexOptimizedFor string) int32 {
...
-			// merging only the large indexes (IVFSQ8 family)
-			// all IVF<same class> indexes are eligible for merging
-			//
-			// try to seek the pointer in the vecIndexes list to a point after
-			// which all indexes are not eligible for fast merge.
-			if i > finalMergeCandidate {
-				finalMergeCandidate = i
-			}
+		}
+		if !drops && vecIndexes[i].index.isMergeable(cfg.optimizationType) {
```
CascadingRadium: given that we do a recons merge if 
- isNotMergeable or
- error in merge_from

I believe we can condense the two APIs into one? 
Taking float32 index for example.

```go
func (f *faissFloat32Index) mergeFrom(other faissIndex, offset int64) error {
	otherFaissIndex, ok := other.(*faissFloat32Index)
	if !ok {
		return errNotSupported
	}
    if !f.isMergeable {
        return errNotSupported
    }
	return f.idx.MergeFrom(otherFaissIndex.idx, offset)
}
```

```go
// todo: need to detect and handle data drift in a more intelligent way
func (v *vectorIndexOpaque) fastMergeIndexes(trainedIndex faissIndexIVF, cfg *faissIndexConfig,
	drops bool, vecIndexes []*vecIndexInfo, w *FileWriter, closeCh chan struct{}) error {
	// create a faissIndex for merged index using nlist and nprobe from trained index's
	// config and we avoid using the GPU for fast merge since the underlying faiss
	// code doesn't support merge_from API on GPU indexes.
	nprobe, nlist := trainedIndex.ivfParams()
	cfg.nlist = nlist
	cfg.useGPU = false
	targetIdx, err := faissIndexFactory(cfg)
	if err != nil {
		return err
	}
	defer targetIdx.close()
	// cast to IVF index to be able to set the quantizer for the fast merge
	ivfTargetIdx := targetIdx.castIVF()
	if ivfTargetIdx == nil {
		return ErrorFastMergeIndexNotIVF
	}
	err = ivfTargetIdx.setDirectMap(1)
	if err != nil {
		return err
	}
	// setting the same nprobe value in the merged index as the centroid
	// index to ensure that we probe the same number of clusters
	ivfTargetIdx.setNProbe(int32(nprobe))
	// The trained index will be an IVFSQ8 index - but with no vectors in it.
	// The coarse quantizer of that index will be cloned over here
	err = ivfTargetIdx.setQuantizers(trainedIndex)
	if err != nil {
		return err
	}
	var reconsVecs []float32
	reconsAndAddFrom := func(vecIDs []int64, srcIdx faissIndex) error {
		neededReconsLen := len(vecIDs) * cfg.dimension
		if cap(reconsVecs) < neededReconsLen {
			reconsVecs = make([]float32, neededReconsLen)
		}
		reconsVecs = reconsVecs[:neededReconsLen]
		reconsVecs, err = srcIdx.reconstructBatch(vecIDs, reconsVecs)
		if err != nil {
			return err
		}
		vecSet, err := newVectorSet(cfg.dimension, reconsVecs)
		if err != nil {
			return err
		}
		if cfg.indexType == faissBIVFIndex {
			vecSet.binarize()
		}
		// add to target index the reconstructed vectors for the valid vector IDs from the source index.
		err = targetIdx.add(vecSet)
		if err != nil {
			return err
		}
		reconsVecs = reconsVecs[:0]
		return nil
	}

	for _, vi := range vecIndexes {
		if isClosed(closeCh) {
			return seg.ErrClosed
		}
		srcIdx := vi.index
		if drops {
			// if we have drops, then we have to reconstruct and add the vectors one by one from the source index to the target index,
			// since we can't rely on the merge_from API of faiss because of the data drift between the indexes caused by the drops.
			err = reconsAndAddFrom(vi.vecIds, srcIdx)
			if err != nil {
				return err
			}
		} else {
			// if no drops, then we can use merge_from
			if err = ivfTargetIdx.mergeFrom(srcIdx, targetIdx.ntotal()); err != nil {
				// error here indicates that either:
				// -- Index does not support merge_from OR
				// -- some unexpected issue during merge_from.
				// in either case we can fallback to reconstructing and adding the vectors one by one
				// from the source index to the target index as a error handling mechanism.
				err = reconsAndAddFrom(vi.vecIds, srcIdx)
				if err != nil {
					return err
				}
			}
		}
	}
	tempBuf := v.grabBuf(binary.MaxVarintLen64)
	// write the type of the vector index
	n := binary.PutUvarint(tempBuf, uint64(cfg.indexType))
	_, err = w.Write(tempBuf[:n])
	if err != nil {
		return err
	}
	return targetIdx.write(tempBuf, w)
}

```

faiss_vector_index_float32.go
```diff
@@ -40,6 +42,20 @@ func newFaissFloat32Index(idx *faiss.IndexImpl) (index faissIndex, err error) {
 	}, nil
 }
 
+func newFaissFloat32IndexWithConfig(idx *faiss.IndexImpl, cfg *faissIndexConfig) (index faissIndex, err error) {
```
CascadingRadium, line 45: This new constructor feels redundant, especially since it would entail addition of nil checks for cfg down the line. Also we do not need the entire config here since we are only pertinent to the optimization type.

I think we should modify the existing constructor to 

```go
func newFaissFloat32Index(idx *faiss.IndexImpl, optimization string) (index faissIndex, err error) {
	if idx == nil {
		return nil, errNilIndex
	}
	return &faissFloat32Index{
		optimizationType: optimization,
		idx:              idx,
	}, nil
}
```

This allows to: Identify the type of index within the isMergeable() API without needing to pass in the optimization string.

As per my other comment, please use this opt string to calculate minVecsForMerge and use that only, keeping one common point of truth for all index types.

section_faiss_vector_index.go
CascadingRadium: From what i understand the logic you have proposed is like this:

_ALGORITHM FAST MERGE_

```go
Function:
Input  = trained index `T`, `N` vector indexes to merge.
Output = target index `X` with `V` number of vectors (sumof ntotal() for each index in `N`)

- If `T` is absent OR (`T` is !`mergeable` and `X` is !`mergeable`):
    - Create a target index `X`
    - For each index `Y` in `N` to be merged:
         - `X`.Reconstruct-Train-Add_MergeFrom(`Y`)
- Else:
    - Create a target index `X`.
    - Set quantizers of `X` using `T`.
    - For each index `Y` in `N` to be merged:
         - If `Y` is `mergeable` and no vectors are dropped in `N`: 
             - `X`.Native_MergeFrom(`Y`)
         - Else:
             - `X`.Reconstruct-Add_MergeFrom(`Y`)
             
`X`.WriteTo(File)
---
`mergeable` is true for an `index` if and only if:
- `nvecs > 10000` for recall and latency opt
- `nvecs > 1000` for mem and rabitq opt
- `nvecs > 1000` for bivf-flat and bivf-sq8

where `nvecs = index.ntotal()`.
---
```

1. Can you simplify the logic by adding a helper that just returns the mininum nvecs per opt? - `func minVecsForFastMerge(opt) int`. We can use this to only do checks for nvecs everywhere instead of resorting to a combination of optimization type and nvecs. (for brevity).

2. Since the optimization type is constant for a field, we can simply have the following check right?

```go
	opt := indexOptimizedFor
	minVecsRequired := minVecsForFastMerge(opt)
	if trainedIndex != nil {
		isTrainedIndexMergeable := trainedIndex.ntotal() > minVecsRequired
		isFinalIndexMergeable := nvecs > minVecsRequired
		if isTrainedIndexMergeable && isFinalIndexMergeable {
			err := v.fastMergeIndexes(trainedIndex, config, drops, vecIndexes, w, closeCh)
			if err != nil {
				return err
			}
			// free the indexes as we won't need them anymore after the fast merge
			freeReconstructedIndexes(vecIndexes)
			return nil
		}
	}
```

I am assuiming this effectively achieves what you are proposing already. I suggest dropping this quantization check because it complicates an already complex system. We do not want to have a check for 3 different things which all point to the same thing.

section_faiss_vector_index.go
Thejas-bhat: yeah nice one, let me refactor this. thanks

section_faiss_vector_index.go
Likith101: We hit this condition and error out when the optimization type is RaBitQ and the number of vectors is less than 1000.

That is because when the number of vectors is less than 1000, we always use a "Flat" index without exceptions. So the cast to IVF will fail.

I am assuming this is the case for the other optimization types that are not bound by the ivfSq8Threshold like memory-efficient and the bivf ones. We need a better check to determine if fastmerge is possible on line 577.  Perhaps in the quantizationType check, we need to add an additional case where if nvecs is less than 1000, return "Flat"

Likith101 (dismissed):

section_faiss_vector_index.go
CascadingRadium: hey, can you look at my related comment regarding the algorithm? I have mentioned a similar point there, regarding the mergeability condition being nvecs > 1000

faiss_vector_index_float32.go
Thejas-bhat: the reason why i added this new type of constructor is because we don't have to touch the existing constructor calls that are there, since right now its pertinent to the fast merge path. 

i kept the pointer to config as part of this struct because it feels like something that might be useful in the future.

i definitely agree that we need a point of truth to compare the trained index and the final merged index to say whether we should hit the fast merge path or not - however its better to have the quantization type as the metric of comparison since there's a strict distinction. having a vector count based comparison won't be an accurate way doing things - if the trained index is IVFSQ8 (which is what's going to be the usual case for default optimizations) then its mergeable and in the case where the final merged index nvecs < 10000 (also mergeable since its > 1000) then we're gonna build an IVFFlat index and this discrepancy will cause issues. 

or did you mean like return bool as to whether the nvecs is above a certain threshold for an opt as part of the helper func?

section_faiss_vector_index.go
Thejas-bhat: > we need to add an additional case where if nvecs is less than 1000, return "Flat"

yeah we do return the default type to be Flat but we won't be doing a fast merge in that case.

> We hit this condition and error out when the optimization type is RaBitQ and the number of vectors is less than 1000.
That is because when the number of vectors is less than 1000, we always use a "Flat" index without exceptions. So the cast to IVF will fail.

i don't think we do a fast merge when the final merged index is Flat though, like firstly we do it only if the trained index type is IVF (otherwise we fallback to the reconstruction+re-training way of things). and the second part of the 577 check ensures safe operations if the trained index was of type RabitQ, but the final merged index returns type to be flat and because of the mismatch we would fall back to the reconstruction+re-training.

faiss_vector_index_float32.go
CascadingRadium: I meant a helper that does this:

```go
func minVecsForFastMerge(opt) int
``` 

Basically, if the trained index is mergeable - it means it must have atleast 10K vectors. If  `final merged index nvecs < 10000`, it would mean that the final index created would be IVF flat and hence we should not hit the fast merge API. This is the condition specified in this part of the algorithm:

```
- If `T` is present AND (`T` is `mergeable` and `X` is `mergeable`):
-   call fastMergeIndex()
```

This means that the trained index is ivf sq8 and final index also should be ivfsq8. and if there is any mismatch there we just do reconstruct-train-merge.

section_faiss_vector_index.go
CascadingRadium: with the latest patch, the quantizer check passes without any check for nvecs though. 

```go
func quantizationType(opt string, ntotal int64) string {
	if opt == index.IndexIVFRaBitQ {
		return "RaBitQ"
	} 
```
since the opt is `ivf,rabitq`, the `quantizationType` call in L555

```go
if trainedIndex != nil && trainedIndex.quantization() == quantizationType(indexOptimizedFor, int64(nvecs)) {
```
 will succeed even when the nvecs < 1000 causing the error to pop up.

Thejas-bhat: oh yeah my bad, i'll include this in the helper func. 

okay i'll change the helper func to what you've suggested since both of them are achieving the same thing using the same set of params to check, so neither of them complicate the system worse than the other.

section_faiss_vector_index.go
CascadingRadium, line 55: add a const 

```go
// The threshold for number of vectors beyond which we consider 
// using SQ8 quantization for IVF indexes.
ivfSq8Threshold = 10000
// The threshold for number of vectors beyond which we consider 
// using IVF indexes over Flat indexes.
ivfThreshold = 1000
```

- L736

```go
func determineFloat32IndexToUse(nvecs, nlist int, optimizationType string) string {
	if nvecs < ivfThreshold {
		return "Flat"
	}
```

- L723

```go
// determineCentroids determines the number of centroids to use for an IVF index.
func determineCentroids(nvecs int) int {
	var nlist int
	switch {
	case nvecs >= 200000:
		nlist = int(4 * math.Sqrt(float64(nvecs)))
	case nvecs >= ivfThreshold:
```

- L680

```go
func determineBinaryIndexToUse(nvecs, nlist int) string {
	switch {
	case nvecs >= ivfThreshold:
		return fmt.Sprintf("BIVF%d", nlist)
	default:
		return "BFlat"
	}
}
```

section_faiss_vector_index.go
```diff
@@ -567,15 +554,13 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(centroidIndex faissIndexI
...
-	if centroidIndex != nil &&
-		(indexOptimizedFor == index.IndexIVFRaBitQ ||
-			(nvecs >= ivfSq8Threshold && indexType == faissFP32Index)) {
-		err := v.fastMergeIndexes(centroidIndex, dims, metric, indexType, indexOptimizedFor, vecIndexes, w, closeCh)
+	// create the faiss index to hold the merged data, either via fast merge or reconstruction
+	config := newFaissIndexConfig(indexType, indexOptimizedFor, dims, metric, nvecs, determineCentroids(nvecs), useGPU)
+	// if we have a trained index use the fast merge to perform the merge without
+	// re-training and reconstructing the vectors. we perform the fast merge
+	// only if the quantization type matches between the trained index and the final merged index
+	if trainedIndex != nil && trainedIndex.isMergeable() && canFastMerge(indexOptimizedFor, nvecs) {
```
CascadingRadium: This causes a bug:

If opt is bivf-flat, we should support
setQuantizer + reconstruct_add merge.
With this, since trainedIndex is !isMergeable() we always fallback to reconstruct_train_add merge.

Add this

```go
// returns true if the final merged index is compatible with the trained index for fast merge
// based on their vector counts
func canFastMerge(trainedIndex faissIndexIVF, opt string, totalVecs int64) bool {
    if trainedIndex == nil {
       return false
    }
	var minVecsNeeded int64
	switch opt {
	case index.IndexBIVFWithBackingFlat, index.IndexBIVFWithBackingSQ8:
		fallthrough
	case index.IndexOptimizedForMemoryEfficient:
		fallthrough
	case index.IndexIVFRaBitQ:
		minVecsNeeded = ivfThreshold
	case index.IndexOptimizedForRecall, index.IndexOptimizedForLatency:
		minVecsNeeded = ivfSq8Threshold
	default:
		return false
	}
	return totalVecs > minVecsNeeded && trainedIndex.ntotal() > minVecsNeeded
}
```

```go
// if we are not using GPU for merging and the trained index is compatible for fast merge with the merged index
// then we can do a fast merge using faiss's native merge capabilities, which is significantly faster than
// reconstructing and adding the vectors one by one.
if !useGPU && canFastMerge(trainedIndex, indexOptimizedFor, int64(nvecs)) {
	err := v.fastMergeIndexes(trainedIndex, config, drops, vecIndexes, w, closeCh)
	if err != nil {
		return err
	}
	// free the indexes as we won't need them anymore after the fast merge
	freeReconstructedIndexes(vecIndexes)
	return nil
}
```

faiss_vector_index_float32.go
```diff
@@ -146,20 +162,44 @@ func (f *faissFloat32Index) trainAndAdd(trainingData *vectorSet, vecsToAdd *vect
...
 		// for the current index, return an error.
 		return errNotSupported
 	}
 	return f.idx.SetQuantizers(centroidFaissIndex.idx)
 }
 
+func (f *faissFloat32Index) isMergeable() bool {
+	if f.cfg != nil {
+		switch f.cfg.optimizationType {
+		case index.IndexBIVFWithBackingFlat:
```
CascadingRadium: I think only fp32 index opt should be here, like:

```go
func (f *faissFloat32Index) isMergeable() bool {
	if f.cfg != nil {
		switch f.cfg.optimizationType {
		case index.IndexOptimizedForLatency, index.IndexOptimizedForRecall:
			return f.ntotal() > ivfSq8Threshold
		case index.IndexOptimizedForMemoryEfficient, index.IndexIVFRaBitQ:
			return f.ntotal() > ivfThreshold
        default:
            return false
		}
	}
	return false
}

File: faiss_vector_index.go [hunk 21]
```diff
@@ -41,6 +42,8 @@ type faissIndex interface {
 	metricType() int
 	// ntotal returns the total number of vectors currently stored in the index.
 	ntotal() int64
+	// returns whether index is eligible for fast merge
```
CascadingRadium: i think this API need not be exposed in the interface, as it is used only by the  `mergeFrom` API

faiss_vector_index_bivf.go
```diff
@@ -251,11 +268,53 @@ func (b *faissBinaryIndex) trainAndAdd(trainingData *vectorSet, vecsToAdd *vecto
...
+		err = b.binary.SetQuantizers(idx.binary)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
 	return errNotSupported
 }
 
+func (b *faissBinaryIndex) isMergeable() bool {
```
CascadingRadium, line 289:
```go
func (b *faissBinaryIndex) isMergeable() bool {
	if b.cfg != nil {
		switch b.cfg.optimizationType {
		case index.IndexBIVFWithBackingFlat:
             // cannot call merge from API as flat index doesnt 
             // support it for now
             return false
        case index.IndexBIVFWithBackingSQ8:
			return b.ntotal() > ivfThreshold
		}
	}
	return false
}
```

CascadingRadium (commented):
looks good now, few final bugs to iron out.

CascadingRadium (changes requested):

section_faiss_vector_index.go
CascadingRadium: also can you reposition the `canFastMerge` API after `faissIndexFactory` in the file.

faiss_vector_index_bivf.go
Likith101: +1

abhinavdangeti (dismissed):

CascadingRadium (approved):
LGTM!

Pls wait for one more review

abhinavdangeti (approved):

abhinavdangeti:
`v17.1.0`

PR #394, opened 2026-04-09
Simplify the implementation of vector search filter logic

Thejas-bhat:
- we don't ideally need two different functions in the interface which can be consolidated as part of the go-faiss's `Selector` interface, which is actually a parameter to these functions
- needs to be backported to v16.x as well https://github.com/blevesearch/zapx/pull/396

### DIFF
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -46,11 +46,7 @@ type faissIndex interface {
 	// performs a search on the index using the provided query vector and parameters, with an optional
 	// exclude selector to indicate a "blocklist" of indexed vectors to ignore during search.
 	// It returns the distances and corresponding vector IDs of the top k results.
-	searchWithoutIDs(qVector *vectorSet, k int64, selector faiss.Selector, params json.RawMessage) ([]float32, []int64, error)
-	// performs a search on the index using the provided query vector and parameters, with a required
-	// include selector to indicate an "allowlist" of indexed vectors to consider during search.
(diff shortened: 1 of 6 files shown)


abhinavdangeti (approved):

abhinavdangeti:
`v17.0.9`

CascadingRadium:
Hey @Thejas-bhat, I do not understand the need or the urgency of this PR. 
I am seeing some issues here and I am wondering why this PR was merged in with this haste without a proper review from the team, given that the PR was marked as ready for review a few hours back.

1. When we have a search request coming in on a segment with no exclude bitmap (which happens when there are no deleted vectors in the segment) - the exclude bitmap is `nil` and hence `selector` too will be nil. In this case, if the `params` is not `nil` and `selector` is `nil` (for float32 indices), we just call the go-faiss `SearchWithSelector` API which will error out. This is orthogonal to the earlier behavior where we handled this case well, where we created a valid `SearchParams`.

2. When using for BIVF indices, in the case where `selector` is not nil and `params` is nil, we call `SearchWithSelector` where simply ignore the selector and call `b.Search(xb, k)`

3. For BIVF incides again: if `selector` is nil (no exclusions) and `params` is not `nil`, we simply error out.

cc @abhinavdangeti

abhinavdangeti:
Good spot @CascadingRadium  👍🏼 , let's get these addressed soon.

PR #395, opened 2026-04-13
MB-71383: Add callbacks for GPU errors

capemox:
GPU error callbacks are passed down from bleve through the segment config. These are passed to the GPU indexes.


CascadingRadium:
-2 

hey @capemox, this is very invasive approach for the logging. From what i see, all of this is just to print a few lines in the couchbase logs and increment a counter.
I would suggest a simple counter would be more than enough, we do not want to print the exact reason as to why the clone failed into the logs, a plain counter like `TotGPUCloneFailed` should be sufficient. 
We already have scorch metrics in bleve (look at stats.go) which we poll using prometheus from cbft. Look into reusing the `UpdateFieldStats` API

capemox:
@CascadingRadium Yes, I agree that it's super invasive, but I do think we really need logs to go along with the errors in some fashion. GPUs are notorious for having random failures. We'll discuss further on a more clean way of logging these errors. I'll polish up these PRs to expose the stats through the scorch index stats for now.

CascadingRadium:
Hey @capemox, logging the reason for gpu clone failure could perhaps be looked into later on, i think for now just the stat is good enough. Please ensure to keep it generic, the GPU specific code must not be added in the generic wrapper and must be situated in the `gpu` file added (the struct implementing the interface).

PR #399, opened 2026-04-16
Added Documentation and CMD Line Tooling for BQ

Likith101:


CascadingRadium (dismissed):

abhinavdangeti (approved):

cmd/zap/cmd/vector.go
```diff
@@ -70,21 +70,37 @@ var vectorCmd = &cobra.Command{
...
 		if err != nil {
 			return fmt.Errorf("error reading faiss index from buffer: %v", err)
 		}
-		metricType := vecIndex.MetricType()
-		dims := vecIndex.D()
+		metricType := vecIndexPrimary.MetricType()
+		dims := vecIndexPrimary.D()
+		var indexSizeBacking uint64
+		var vecIndexBacking faiss.Index
+		if vecType == 1 {
```
CascadingRadium, line 89: Should this not be faissBIVFIndex?

cmd/zap/cmd/vector.go
```diff
@@ -70,21 +70,37 @@ var vectorCmd = &cobra.Command{
...
+			// read the length of the serialized backing faiss index
+			indexSizeBacking, n = binary.Uvarint(data[pos : pos+binary.MaxVarintLen64])
+			pos += uint64(n)
+			indexBytesBacking := data[pos : pos+indexSizeBacking]
+			pos += indexSizeBacking
+			// construct the backing faiss index from the buffer
+			vecIndexBacking, err = faiss.ReadIndexFromBuffer(indexBytesBacking, faiss.IOFlagReadOnly)
```
CascadingRadium, line 96: use `faissIOFlagsReadOnly`

cmd/zap/cmd/vector.go
Likith101: we cannot import faissIndexType into this module, so we have to check the value directly

cmd/zap/cmd/vector.go
Likith101: Ditto

CascadingRadium (approved):

PR #400, opened 2026-04-20
MB-71383: Report vector stats for each segment

capemox:
Exposes `UpdateVectorFieldStats` for each segment, which collects all the vector indexes that are either in GPU or CPU memory.
- If the index is not active (cache entry is nil), skip
- If the index is active and is GPU, increment GPU counter
- If the index is active and is in CPU, increment CPU counter


faiss_vector_posting.go
```diff
@@ -334,3 +334,30 @@ func (sb *SegmentBase) UpdateFieldStats(stats segment.FieldStats) {
...
+		if entry == nil {
+			// not in cache — not actively in use, skip
+			continue
+		}
+		gpuIdx, ok := entry.index.(*faissGPUFloat32Index)
+		if !ok {
+			continue
+		}
+		if gpuIdx.isInGPU() {
+			stats.Store("num_gpu_segments_in_gpu_memory", fieldName, 1)
```
abhinavdangeti: `num_segments_in_gpu_mem`*

faiss_vector_posting.go
```diff
@@ -334,3 +334,30 @@ func (sb *SegmentBase) UpdateFieldStats(stats segment.FieldStats) {
...
+		} else {
+			stats.Store("num_gpu_segments_in_cpu_memory", fieldName, 1)
```
abhinavdangeti: `num_segments_in_cpu_mem`*

faiss_vector_index_gpu_float32.go
```diff
@@ -168,6 +168,12 @@ func (f *faissGPUFloat32Index) size() uint64 {
 	return f.cpuIdx.Size()
 }
 
+// isInGPU reports if the index is currently running on the GPU.
+// returns false if the async clone is not yet done or the clone fails.
+func (f *faissGPUFloat32Index) isInGPU() bool {
```
abhinavdangeti: Better name needed :) - perhaps `inGPURam()`?

abhinavdangeti (commented):
Is `stats.Store(..)` the right API - it seems we'll just keep updating it's value to 1?

capemox:
@abhinavdangeti each segment has only 1 vector index per field, and in scorch we merge the stats for all the fields across all the segments, so it should work as expected.

In any case, I'll add unit tests to the bleve PR to make sure this works as we expect.

faiss_vector_posting.go
capemox: @abhinavdangeti We track only the gpu *indexes that are in gpu/cpu mem, so it's better we specify that these are GPU indexes. If we do these for all vector indexes, the num indexes in cpu memory won't represent what we want, which is the number of gpu indexes that are being served from cpu due to some issue.

This is why I think we should specify the name.

faiss_vector_posting.go
capemox: Same as below

faiss_vector_posting.go
abhinavdangeti: Ah, cool 👍🏼

abhinavdangeti (dismissed):

faiss_vector_posting.go
```diff
@@ -334,3 +334,30 @@ func (sb *SegmentBase) UpdateFieldStats(stats segment.FieldStats) {
...
+		if entry == nil {
+			// not in cache — not actively in use, skip
+			continue
+		}
+		gpuIdx, ok := entry.index.(*faissGPUFloat32Index)
+		if !ok {
+			continue
+		}
+		if gpuIdx.inGPURam() {
+			stats.Store("num_gpu_segments_in_gpu_memory", fieldName, 1)
```
CascadingRadium: stat name feels confusing, when we aggregate across segments later on, we will get one stat per field.

keep it `num_vector_indexes_in_gpu` and `num_vector_indexes_in_cpu`

capemox: Will change to `num_gpu_vector_indexes_in_gpu/cpu_memory` instead.

I think we need to specify that's in gpu since we are only tracking this for GPU indexes. CPU indexes won't be tracked at all.

CascadingRadium: Why not do this:

Have the stat enabled for all vector fields. 
For every field the stat becomes valid then. For plain fp32/bq type we can see the 
`num_vector_indexes_in_cpu` stat. 

```go
if gpuIdx, ok := entry.index.(*faissGPUFloat32Index); ok && gpuIdx.inGPURam() {
	stats.Store("num_vector_indexes_in_gpu", fieldName, 1) 
} else {
	stats.Store("num_vector_indexes_in_cpu", fieldName, 1)
}
```

This will be useful as we can observe the `liveness` of the index. If the stat is 0 then the cache is not loaded.

capemox: Sure, since we're ok to expose it for vector fields.

CascadingRadium: simplified the suggestion a bit to avoid line duplication

faiss_vector_posting.go
```diff
@@ -334,3 +334,30 @@ func (sb *SegmentBase) UpdateFieldStats(stats segment.FieldStats) {
...
+		if pos == 0 {
+			continue
+		}
+		fieldID := sb.fieldsMap[fieldName] - 1
+		entry := sb.vecIndexCache.getEntry(fieldID)
+		if entry == nil {
+			// not in cache — not actively in use, skip
+			continue
+		}
+		gpuIdx, ok := entry.index.(*faissGPUFloat32Index)
```
CascadingRadium: One issue here, we let go of the read lock when we return from `getEntry`. The `cleanup` API might race and close the `entry` object returned while we perform the check below. Maybe we can modify the `getEntry` API to return the necessary information there? Rename to `isLive` maybe?

```go
// returns
// - true, true is index is in cache and loaded in gpu
// - true, false if index is in cache but not in gpu
// - false, false if index is not in cache and not in gpu
func (vc *vectorIndexCache) isLive(fieldID uint16) (bool, bool)
```

faiss_vector_cache.go
```diff
@@ -228,15 +228,29 @@ func (vc *vectorIndexCache) decRef(fieldID uint16) {
...
 	defer vc.m.RUnlock()
 	if vc.isClosed {
-		return nil
+		return vectorIndexNotCached
+	}
+	entry, ok := vc.cache[fieldID]
+	if !ok {
+		return vectorIndexNotCached
+	}
+	if gpuIdx, ok := entry.index.(*faissGPUFloat32Index); ok && gpuIdx.inGPURam() {
```
CascadingRadium: One last comment. Add a `faissIndexGPU` interface which has the  `inGPU()` API. That way we can do an interface check here instead of hard type assert. When we add the bivf-GPU impl of the interface, we do not need to revisit this again then

faiss_vector_cache.go
```diff
@@ -228,15 +228,29 @@ func (vc *vectorIndexCache) decRef(fieldID uint16) {
 	vc.m.RUnlock()
 }
 
-// getEntry returns the cache entry for the given fieldID, or nil if the field
-// is not currently cached (i.e. has never been queried or has been evicted).
-func (vc *vectorIndexCache) getEntry(fieldID uint16) *cacheEntry {
+type vectorIndexLocation uint8
+
+const (
```
CascadingRadium, line 234: add some commentary

CascadingRadium (commented):
LGTM. A few last comments for the sake of extensibility. Nice work 👍🏻

CascadingRadium (dismissed):
Please update PR description and title before merging.

abhinavdangeti (approved):

PR #401, opened 2026-04-20
MB-71516: Fix result inconsistencies when exhausting binary indices

CascadingRadium:
- When we exhaust a vector index - when `k` > `N` (where `N` represents the number of "valid"
  vectors i.e. those vectors that have not been excluded using the bitmap), FAISS pads the
  output `labels` with `-1` and the corresponding `distances` with `±Inf`, depending on the
  distance metric used.
- We currently process the current iteration's output only until we see the first `-1` and
  perform an early exit, assuming that the rest of the output will be `-1` as well following
  FAISS's padding logic.
- With binary indices in play, due to the inherent nature of the QuickSelect algorithm used to
  pick the top K IDs, the final ordering of the result set output will be random, resulting in
  the previous padding assumption falling through, since we may yet have valid results beyond
  the first `-1`.
- Fix this issue by always processing the complete result set of the current iteration,
  filtering out any `-1` entries regardless of their position, rather than stopping at the
  first `-1` encountered. This ensures all valid results are always captured.

### DIFF
--- a/faiss_vector_wrapper.go
+++ b/faiss_vector_wrapper.go
@@ -331,13 +331,13 @@ func (v *vectorIndexWrapper) docSearch(k int64, numDocs uint64,
 		}
 		// process the retrieved ids and scores, getting the corresponding docIDs
 		// for each vector id retrieved, and storing the best score for each unique docID
-		// the moment we see a -1 for a vector id, we stop processing further since
-		// it indicates there are no more vectors to be retrieved and break out of the loop
-		// by setting the exhausted flag
 		for i, vecID := range labels {
+			// a vecID of -1 indicates that all valid vectors in the index have been exhausted,
+			// so we set the flag to prevent further iterations. However, the current iteration
+			// may still contain valid results, so we process them before stopping.
 			if vecID == -1 {
 				exhausted = true
-				break
+				continue
 			}
 			docID, exists := v.getDocIDForVectorID(vecID)
 			if !exists {
--- a/faiss_vector_test.go
+++ b/faiss_vector_test.go
@@ -156,6 +156,18 @@ func newStubFieldVec(name string, vector []float32, d int, metric string, fieldO
 	}
 }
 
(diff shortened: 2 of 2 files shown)


abhinavdangeti (commented):
Let's add a simple unit test in-line here.

CascadingRadium:
done

faiss_vector_wrapper.go
```diff
@@ -331,13 +331,13 @@ func (v *vectorIndexWrapper) docSearch(k int64, numDocs uint64,
...
 		// for each vector id retrieved, and storing the best score for each unique docID
-		// the moment we see a -1 for a vector id, we stop processing further since
-		// it indicates there are no more vectors to be retrieved and break out of the loop
-		// by setting the exhausted flag
 		for i, vecID := range labels {
+			// a vecID of -1 indicates that all valid vectors in the index have been exhausted,
+			// so we set the flag to prevent further iterations. However, the current iteration
+			// may still contain valid results, so we process them before stopping.
 			if vecID == -1 {
 				exhausted = true
```
capemox, line 339: If we're doing the exhaustive search, we don't need this `exhausted` flag anymore.

This looks like a change that will affect non-binary indexes as well, will there be a loss in perf in non-binary indexes because of no early exit?

capemox (changes requested):

faiss_vector_wrapper.go
CascadingRadium: no we need this `exhausted` flag to prevent future iterations. This patch only ensures that we process the complete result set of the current iteration.

CascadingRadium: perf impact will be negligible as we will loop until end of current result set which will be bounded by the `k` value.

capemox (approved):

Likith101 (approved):

abhinavdangeti (approved):

PR #403, opened 2026-04-22
make the `faissIndex` interface and implementation more modular

Thejas-bhat:
- update the faissBinaryIndex to remove redundant code from the section_faiss_vector_index.go and perform the operations in a more modular manner.
- update the faissIndex interface to have a write API so that along with serialization it'll write out the content to the FileWriter in a format it's designed for
- when RabitQ is enabled, all faiss indexes having vector count > 1000 have the same quantization check. so the fastmerge condition check is updated to account this as well

### DIFF
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -96,14 +96,6 @@ type faissIndexIVF interface {
 	mergeFrom(other faissIndex, offset int64) error
 }
 
-// Interface for SQ-specific operations on Faiss vector indices.
-type faissIndexSQ interface {
-	// trains the SQ index on the provided training data and adds the vectors to
-	// the trained index. The training step performs quantization of the vector space,
-	// which enables efficient storage and search of high-dimensional vectors.
-	trainAndAdd(trainingData *vectorSet, vecsToAdd *vectorSet) error
-}
-
 // faissIndexGPU is implemented by any index type that can reside in GPU memory.
 type faissIndexGPU interface {
 	// inGPURam reports whether the index is currently loaded in GPU memory.
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -57,9 +57,6 @@ type faissIndex interface {
 	// returns the underlying IVF index if this is an IVF index,
 	// and a boolean indicating whether the cast was successful.
 	castIVF() faissIndexIVF
-	// returns the underlying SQ index if this is an SQ index,
-	// and a boolean indicating whether the cast was successful.
-	castSQ() faissIndexSQ
 }
 
 // Interface for IVF-specific operations on Faiss vector indices.
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -46,9 +46,9 @@ type faissIndex interface {
 	// performs a search on the index using the provided query vector and and retrieves the top K nearest neighbors.
 	// Optional search constraints can be applied using the selector and additional search parameters.
 	search(qVector *vectorSet, k int64, selector faiss.Selector, params json.RawMessage) ([]float32, []int64, error)
-	// serializes the index into a byte slice,
-	// which can be stored or transmitted.
-	serialize() ([]byte, error)
+	// write out the index content into the provide fileWriter using a reusable buffer
+	// returns any error encountered during the write process.
+	write(buf []byte, w *FileWriter) error
 	// returns the size of the index in bytes.
 	size() uint64
 	// -----------------------------------------------------------------
--- a/faiss_vector_index_bivf.go
+++ b/faiss_vector_index_bivf.go
@@ -18,6 +18,7 @@
 package zap
 
 import (
+	"encoding/binary"
 	"encoding/json"
 
 	faiss "github.com/blevesearch/go-faiss"
@@ -33,7 +34,7 @@ type faissBinaryIndex struct {
 
 func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl) (index faissIndex, err error) {
 	// the binary index cannot be nil, but the backing index can be nil, depending on the use case.
-	if binary == nil {
+	if binary == nil || backing == nil {
 		return nil, errNilIndex
 	}
 	return &faissBinaryIndex{
@@ -43,15 +44,17 @@ func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl
 }
 
 func (b *faissBinaryIndex) add(vecs *vectorSet) error {
-	// add the binaryData from the vectorSet to the binary index
+	// add float data to backing index and the binary data to binary index
+	err := b.backing.Add(vecs.floatData)
+	if err != nil {
+		return err
+	}
 	return b.binary.Add(vecs.binaryData)
 }
 
 func (b *faissBinaryIndex) close() {
 	b.binary.Close()
-	if b.backing != nil {
-		b.backing.Close()
-	}
+	b.backing.Close()
 }
 
 func (b *faissBinaryIndex) dim() int {
@@ -59,10 +62,7 @@ func (b *faissBinaryIndex) dim() int {
 }
 
 func (b *faissBinaryIndex) metricType() int {
-	if b.backing != nil {
-		return b.backing.MetricType()
-	}
-	return b.binary.MetricType()
+	return b.backing.MetricType()
 }
 
 func (b *faissBinaryIndex) ntotal() int64 {
@@ -70,8 +70,8 @@ func (b *faissBinaryIndex) ntotal() int64 {
 }
 
 func (b *faissBinaryIndex) reconstructBatch(vecIDs []int64, prealloc []float32) ([]float32, error) {
-	// binary indexes do not support reconstruction
-	return nil, errNotSupported
+	// reconstruct vectors from backing index
+	return b.backing.ReconstructBatch(vecIDs, prealloc)
 }
 
 func (b *faissBinaryIndex) search(qVector *vectorSet, k int64, selector faiss.Selector, params json.RawMessage) ([]float32, []int64, error) {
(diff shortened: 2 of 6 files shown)


faiss_vector_index_bivf.go
```diff
@@ -43,35 +44,34 @@ func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl
...
+	}
 	return b.binary.Add(vecs.binaryData)
 }
 
 func (b *faissBinaryIndex) close() {
 	b.binary.Close()
-	if b.backing != nil {
-		b.backing.Close()
-	}
+	b.backing.Close()
```
abhinavdangeti, line 57: As per comment in line 36 above, if backing can be nil - shouldn't we check before dereferencing?

faiss_vector_index_bivf.go
```diff
@@ -43,35 +44,34 @@ func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl
...
 func (b *faissBinaryIndex) dim() int {
 	return b.binary.D()
 }
 
 func (b *faissBinaryIndex) metricType() int {
-	if b.backing != nil {
-		return b.backing.MetricType()
-	}
-	return b.binary.MetricType()
+	return b.backing.MetricType()
```
abhinavdangeti, line 65: Ditto^

faiss_vector_index_bivf.go
```diff
@@ -43,35 +44,34 @@ func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl
...
 
 func (b *faissBinaryIndex) ntotal() int64 {
 	return b.binary.Ntotal()
 }
 
 func (b *faissBinaryIndex) reconstructBatch(vecIDs []int64, prealloc []float32) ([]float32, error) {
-	// binary indexes do not support reconstruction
-	return nil, errNotSupported
+	// reconstruct vectors from backing index
+	return b.backing.ReconstructBatch(vecIDs, prealloc)
```
abhinavdangeti, line 74: Based on this and the other APIs it seems backing _must not_ be `nil` for the `faissBinaryIndex` ?

abhinavdangeti (commented):
Code positioning looks great 👍🏼 , but I have some comments about `faissBinaryIndex`'s backing index.

faiss_vector_index_bivf.go
Thejas-bhat: yeah so since we're erring out during object creation time we don't need the nil checks - i've updated the comment

abhinavdangeti (dismissed):
👌🏼

faiss_vector_index_bivf.go
```diff
@@ -222,15 +238,24 @@ func (b *faissBinaryIndex) setNProbe(nprobe int32) {
 }
 
 func (b *faissBinaryIndex) trainAndAdd(trainingData *vectorSet, vecsToAdd *vectorSet) error {
-	err := b.binary.Train(trainingData.binaryData)
+	// train the backing index with the floatData
+	var err error
+	if b.backing != nil && b.backing.IsSQIndex() {
```
CascadingRadium: can omit the `if b.backing != nil` check here for consistency

section_faiss_vector_index.go
```diff
@@ -531,11 +555,9 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(centroidIndex faissIndexI
 		return nil
 	}
 	// Fast Merge Path:
-	// hardcoded - refactor later
 	// fast merge only applicable for:
 	// - IVFSQ8 indexes beyond a certain size threshold.
 	// - the centroid index is available.
-	// - the index type is faissFP32Index.
 	if nvecs >= ivfSq8Threshold && centroidIndex != nil && indexType == faissFP32Index {
```
CascadingRadium: should we check for rabitQ here?

section_faiss_vector_index.go
```diff
@@ -349,7 +359,7 @@ func (v *vectorIndexOpaque) fastMergeIndexes(centroidIndex faissIndexIVF, dims,
 		vecCount := len(vecIndexes[i].vecIds)
 		if vecCount == 0 {
 			continue
-		} else if vecCount >= ivfSq8Threshold {
+		} else if vecIndexes[i].indexOptimizedFor == index.IndexIVFRaBitQ || vecCount >= ivfSq8Threshold {
```
CascadingRadium: why do we need this check?

Thejas-bhat: yeah this check needs to be better - basically all the ivf indexes are eligible for fast merge when using rabitq, since they're are of the same quantization and also recons when vec count < 10k will be quite lossy.

@Likith101 are you ensuring on the bleve side that we enable fast merge when using rabitQ? or is this a user responsibility for now?

section_faiss_vector_index.go
Thejas-bhat: hmm yeah i think we can keep an addtional check here as well, since even when the count < sq8 threshold and the child indexes are ivf indexes we shouldn't be reconsrtucting in that case

section_faiss_vector_index.go
CascadingRadium: I think we allow rabitQ optimization on vector fields with a non-fastmerge index mapping.

abhinavdangeti: Nope, not gonna happen - if the user chooses rabitQ, fast merge will be forced.

CascadingRadium (dismissed):
Good PR.

One small cleanup would be to remove the now vestigial castSQ API.

Thejas-bhat:
yeah that makes sense, if in future we need a way to get the backing index for some reason it should ideally be part of a binaryIndex specific interface perhaps

CascadingRadium:
missed the castSQ in faissIndexBinary file

section_faiss_vector_index.go
```diff
@@ -651,38 +653,44 @@ func makeFaissIndex(vecs *vectorSet, config *faissIndexConfig, w *FileWriter) ([
...
 		// we have sequential vector IDs starting from 0 to N-1.
 		err = ivfIndex.setDirectMap(1)
 		if err != nil {
-			return nil, err
+			return err
 		}
 		// calculate nprobe using a heuristic.
 		nprobe := calculateNprobe(config.nlist, config.optimizationType)
 		ivfIndex.setNProbe(nprobe)
-	} else if sqIndex := index.castSQ(); sqIndex != nil {
```
capemox, line 667: It looks like we've removed the arm for an SQ index, will this cause any issues?

faiss_vector_index.go
capemox: I think this interface can be removed now that `castSQ` is removed right?

capemox (changes requested):

faiss_vector_index.go
Thejas-bhat: yeah true

section_faiss_vector_index.go
Thejas-bhat: i dont think it should. since this is a backing index logic, we have shifted it to the bivf file. when the binaryIVF index is being merged, the first block would be it and inside the bivf file we would train+add stuff

faiss_vector_index_bivf.go
```diff
@@ -85,36 +85,59 @@ func (b *faissBinaryIndex) search(qVector *vectorSet, k int64, selector faiss.Se
...
+	_, err = w.Write(buf[:n])
+	if err != nil {
+		return err
+	}
+
+	backingBytes = w.process(backingBytes)
+	_, err = w.Write(backingBytes)
+	if err != nil {
+		return err
+	}
```
Likith101, line 118: Write the length after processing since it can change during the process call

faiss_vector_index_bivf.go
```diff
@@ -85,36 +85,59 @@ func (b *faissBinaryIndex) search(qVector *vectorSet, k int64, selector faiss.Se
...
+	// write the length of the serialized vector index bytes
+	n = binary.PutUvarint(buf, uint64(len(binaryBytes)))
+	_, err = w.Write(buf[:n])
+	if err != nil {
+		return err
+	}
+
+	binaryBytes = w.process(binaryBytes)
+	_, err = w.Write(binaryBytes)
```
Likith101, line 133: Ditto

faiss_vector_index_float32.go
```diff
@@ -67,8 +68,25 @@ func (f *faissFloat32Index) search(qVector *vectorSet, k int64, selector faiss.S
...
+	}
+
+	// write the length of the serialized vector index bytes
+	n := binary.PutUvarint(buf, uint64(len(idxBytes)))
+	_, err = w.Write(buf[:n])
+	if err != nil {
+		return err
+	}
+
+	idxBytes = w.process(idxBytes)
```
Likith101: Same again

faiss_vector_index_gpu_float32.go
```diff
@@ -160,8 +161,25 @@ func (f *faissGPUFloat32Index) search(qVector *vectorSet, k int64, selector fais
...
+	}
+
+	// write the length of the serialized vector index bytes
+	n := binary.PutUvarint(buf, uint64(len(idxBytes)))
+	_, err = w.Write(buf[:n])
+	if err != nil {
+		return err
+	}
+
+	idxBytes = w.process(idxBytes)
```
Likith101: Ditto

section_faiss_vector_index.go
```diff
@@ -349,7 +363,9 @@ func (v *vectorIndexOpaque) fastMergeIndexes(centroidIndex faissIndexIVF, dims,
 		vecCount := len(vecIndexes[i].vecIds)
 		if vecCount == 0 {
 			continue
-		} else if vecCount >= ivfSq8Threshold {
+		} else if (vecIndexes[i].indexOptimizedFor == index.IndexIVFRaBitQ &&
+			vecIndexes[i].index.castIVF() != nil) ||
+			vecCount >= ivfSq8Threshold {
```
Likith101, line 368: If the index optimization type is IVFRaBitQ, why do we need to explicitly check castIVF? Shouldn't it always pass this second check?

Likith101 (changes requested):

section_faiss_vector_index.go
Thejas-bhat: we do create flat index even if the optimization is rabitQ - like when the vector count is < 1000. so we'll have to qualify only the ivf type indexes.

faiss_vector_index_bivf.go
Thejas-bhat: yeah my bad on that, thanks

abhinavdangeti (approved):

CascadingRadium (approved):

capemox (approved):

Likith101 (approved):

PR #407, opened 2026-05-07
MB-71796: Fix crash when using GPU Index Class

CascadingRadium:
- Add a check to prevent unsupported vector index classes from being cloned to the GPU.
  Earlier we could clone to GPU for the `IVF,RaBitQ` class and crash the process.
- Add a check to prevent `IVF,Flat` indexes from being cloned to GPU as we do not expect
  to have much speedup from this vector classes as we create them for <10000 vectors.
- Fix bug where we invalidate fast merge operations due to boundary condition mismatch.
- Cleanup code.

### DIFF
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -25,9 +25,9 @@ import (
 )
 
 var (
-	errNilConfig    error = errors.New("faiss index config is nil")
-	errNilIndex     error = errors.New("faiss index is nil")
-	errNotSupported error = errors.New("operation not supported")
+	errInvalidOptimizationString error = errors.New("faiss index optimization string is invalid")
+	errNilIndex                  error = errors.New("faiss index is nil")
+	errNotSupported              error = errors.New("operation not supported")
 )
 
 // Abstract interface for Faiss vector indices, which are returned by the go-faiss library.
--- a/faiss_vector_index.go
+++ b/faiss_vector_index.go
@@ -25,10 +25,38 @@ import (
 )
 
 var (
-	errInvalidOptimizationString error = errors.New("faiss index optimization string is invalid")
-	errNilIndex                  error = errors.New("faiss index is nil")
-	errNotSupported              error = errors.New("operation not supported")
+	errNilIndex     error = errors.New("faiss index is nil")
+	errNilParams    error = errors.New("faiss index params is nil")
+	errNotSupported error = errors.New("operation not supported")
 )
+
+// -------------------------------------------
+// Parameters for constructing a Faiss index
+// -------------------------------------------
+// faissIndexParams holds the construction-time parameters required by a
+// faissIndex implementation that cannot be derived from the underlying
+// faiss index objects themselves (e.g. dimension and metric type are
+// already available on the faiss index).
+type faissIndexParams struct {
+	// optimization is the index optimization type string (e.g. latency,
+	// recall, memory-efficient, BIVF flavours).
+	optimization string
+	// numVecs is the number of vectors expected to populate the index.
+	// It is needed at construction time because the underlying faiss
+	// index reports zero vectors until trainAndAdd/add is called, but
+	// some construction-time decisions (e.g. whether to clone to GPU)
+	// depend on the eventual vector count.
+	numVecs int
+}
+
+// newFaissIndexParams constructs a faissIndexParams with the given optimization
+// type and expected vector count.
+func newFaissIndexParams(optimization string, numVecs int) *faissIndexParams {
+	return &faissIndexParams{
+		optimization: optimization,
+		numVecs:      numVecs,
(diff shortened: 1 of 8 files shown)


capemox (dismissed):

faiss_vector_index_float32.go
```diff
@@ -29,30 +29,20 @@ import (
 // Faiss Float32 Index
 // ---------------------------------
 type faissFloat32Index struct {
-	cfg *faissIndexConfig
-	idx *faiss.IndexImpl
+	idx          *faiss.IndexImpl
+	optimization string
```
Thejas-bhat: i dont think having such a specific piece of info like optimization type as part of the struct over is a good idea. having a config makes more sense from a long term perspective when we might want to do something more.

Thejas-bhat (changes requested):

faiss_vector_index_gpu_float32.go
```diff
@@ -60,13 +64,18 @@ type faissGPUFloat32Index struct {
 // newFaissGPUFloat32Index creates a GPU-backed float32 index. The GPU clone is
 // always performed asynchronously; search falls back to CPU until it
 // completes. All other GPU-operating methods block on doneCh before proceeding.
-func newFaissGPUFloat32Index(cpuIdx *faiss.IndexImpl) (faissIndex, error) {
+func newFaissGPUFloat32Index(cpuIdx *faiss.IndexImpl, optimization string, nvecs int) (faissIndex, error) {
```
Thejas-bhat: cpu and gpu indexes won't be needing the exact same set of params to do their operations, wouldn't having a config based approach be better in general? it just standardises the constructor signatures. 
having a constructor() and constructorWithConfig() is also a common pattern we have in blevesearch that helps in not forcing a config for cases where it might not be necessary. that can be collapsed into a single constructor signature ofc where if the config is nil, we'll just init with an empty map

CascadingRadium: fixing.

faiss_vector_index_float32.go
CascadingRadium: ack

faiss_vector_index_gpu_float32.go
CascadingRadium: the main reason why i removed the withConfig API was that there were too many details that can either be derived by the faiss index itself or were irrelevant.
The main `faissIndexConfig` API was meant to be used for the `faissIndexFactory` API only, so using the same struct for the object construction would be wrong as we would be exposing fields like `useGPU, metricType` etc. I have fixed this in the latest patch by having a new struct with only relevant details instead - that can be created in the query/read path as well.

faiss_vector_index.go
```diff
@@ -25,11 +25,39 @@ import (
...
 )
 
+// -------------------------------------------
+// Parameters for constructing a Faiss index
+// -------------------------------------------
+// faissIndexParams holds the construction-time parameters required by a
+// faissIndex implementation that cannot be derived from the underlying
+// faiss index objects themselves (e.g. dimension and metric type are
+// already available on the faiss index).
+type faissIndexParams struct {
```
Thejas-bhat, line 40: why is the type faissIndexConfig not reusable while using in the constructor though? its anyways used to build the vector indexes while writing out and while merging, we might as well pass the resuse the same config to the constructor?

faiss_vector_index_gpu_float32.go
Thejas-bhat: i dont think the type faissIndexConfig is exposing anything out of the index domain though right? so it still feels reusable?

CascadingRadium: They model two different lifecycle stages: 
- `faissIndexConfig` is the input to `faissIndexFactory` and describes how to build
   an index (needs `indexType`, `nlist`, `useGPU`, `metricType`, etc.).
- The the wrapper constructors (`newFaissFloat32Index`, `newFaissBinaryIndex`, ...)
  only wrap an already-existing faiss object — at that point
  `dimension`/`metricType` are derivable from the  `faiss` object itself, and
  `nlist`/`useGPU`/`indexType` are write-time-only decisions that are meaningless 
   on the read path. 
- Reusing `faissIndexConfig` would force the query/cache/merge-load paths 
  to fabricate fake values for those fields and re-expose write-path concerns 
- `faissIndexParams` keeps just the two things the wrapper genuinely needs and 
   can't recover from the faiss object (`optimization` and `numVecs`), so it's 
   constructible from both write and read paths without lying.

Thejas-bhat: yeah cool, sounds good

Thejas-bhat (approved):

Likith101 (approved):

PR #411, opened 2026-05-15
MB-71936: Prevent GC of vector index when filecallbacks is enabled

Likith101:
- Added new constructors for query path that keep a reference of the original index bytes slices preventing gc
 - Added filecallback related changes to centroid indexes

### DIFF
--- a/centroid_index.go
+++ b/centroid_index.go
@@ -70,7 +70,10 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
 	params := newFaissIndexParams(optStr, int(numVecs))
 
 	// todo: might wanna use the vector cache here, early tests didn't show a big diff
-	fIndexBytes := sb.mem[pos : pos+indexSize]
+	fIndexBytes, err := sb.fileReader.process(sb.mem[pos : pos+indexSize])
+	if err != nil {
+		return nil, err
+	}
 	pos += indexSize
 
 	if faissIndexType(indexType) == faissBIVFIndex {
--- a/centroid_index.go
+++ b/centroid_index.go
@@ -52,14 +52,9 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
 	pos += uint64(n)
 
 	// length of the vector to docID map
-	_, n = binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
+	mapLen, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
 	pos += uint64(n)
-
-	// vector to docID mapping
-	for i := 0; i < int(numVecs); i++ {
-		_, n = binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-		pos += uint64(n)
-	}
+	pos += mapLen
 
 	// type of index
 	indexType, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
--- a/centroid_index.go
+++ b/centroid_index.go
@@ -68,8 +68,12 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
 	indexSize, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
 	pos += uint64(n)
 
+	ioFlags := faissIOFlags
+	if sb.fileReader.id != DefaultFileCallbackId {
+		ioFlags = faissIOFlagsFileCallbacks
+	}
 	// todo: might wanna use the vector cache here, early tests didn't show a big diff
-	faissIndex, err := faiss.ReadIndexFromBuffer(sb.mem[pos:pos+indexSize], faissIOFlags)
+	faissIndex, err := faiss.ReadIndexFromBuffer(sb.mem[pos:pos+int(indexSize)], ioFlags)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


file_callbacks.go
```diff
@@ -38,6 +38,8 @@ import (
 // An example implementation using AES-GCM encryption is provided in
 // file_callbacks_test.go within initFileCallbacks().
 
+const DefaultFileCallbackId = ""
```
CascadingRadium, line 42: `DefaultFileCallbackID = ""`

Can you add a comment here detailing what it is.

faiss_vector_io_flags_unix.go
```diff
@@ -20,6 +20,7 @@ package zap
 import faiss "github.com/blevesearch/go-faiss"
 
 const (
-	faissIOFlags         = faiss.IOFlagReadMmap | faiss.IOFlagSkipPrefetch
-	faissIOFlagsReadOnly = faissIOFlags | faiss.IOFlagReadOnly
+	faissIOFlags              = faiss.IOFlagReadMmap | faiss.IOFlagSkipPrefetch
+	faissIOFlagsReadOnly      = faissIOFlags | faiss.IOFlagReadOnly
+	faissIOFlagsFileCallbacks = faiss.IOFlagReadOnly
```
capemox: Might also want to add a small comment here

faiss_vector_io_flags_win.go
```diff
@@ -20,6 +20,7 @@ package zap
 import faiss "github.com/blevesearch/go-faiss"
 
 const (
-	faissIOFlags         = faiss.IOFlagReadOnly
-	faissIOFlagsReadOnly = faiss.IOFlagReadOnly
+	faissIOFlags              = faiss.IOFlagReadOnly
+	faissIOFlagsReadOnly      = faiss.IOFlagReadOnly
+	faissIOFlagsFileCallbacks = faiss.IOFlagReadOnly
```
capemox: ditto

capemox (dismissed):

centroid_index.go
```diff
@@ -68,22 +67,17 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
...
 	pos += uint64(n)
 
+	params := newFaissIndexParams(optStr, int(numVecs))
+
 	// todo: might wanna use the vector cache here, early tests didn't show a big diff
-	faissIndex, err := faiss.ReadIndexFromBuffer(sb.mem[pos:pos+indexSize], faissIOFlags)
-	if err != nil {
-		return nil, err
-	}
+	fIndexBytes := sb.mem[pos : pos+indexSize]
```
CascadingRadium: shouldn't we be calling the process API here ?

faiss_vector_index_bivf.go
```diff
@@ -49,6 +54,34 @@ func newFaissBinaryIndex(binary *faiss.BinaryIndexImpl, backing *faiss.IndexImpl
...
+func newFaissBinaryIndexFromBytes(bIndexBytes, fIndexBytes []byte, params *faissIndexParams) (faissIndex, error) {
+	if bIndexBytes == nil || fIndexBytes == nil {
+		return nil, errNilIndex
+	}
+
+	if params == nil {
+		return nil, errNilParams
+	}
+
+	backing, err := faiss.ReadIndexFromBuffer(bIndexBytes, faissIOFlagsReadOnly)
```
CascadingRadium, line 66: should this be `faissIOFlags` or `faissIOFlagsReadOnly`? Before we were mixing it up. Just to confirm.

faiss_vector_index_float32.go
```diff
@@ -46,12 +48,30 @@ func newFaissFloat32Index(idx *faiss.IndexImpl, params *faissIndexParams) (faiss
 	}, nil
 }
 
+func newFaissFloat32IndexFromBytes(idxBytes []byte, params *faissIndexParams) (faissIndex, error) {
+	if idxBytes == nil {
+		return nil, errNilIndex
+	}
+
```
CascadingRadium, line 55: missing check for `params == nil`

faiss_vector_index_gpu_float32.go
```diff
@@ -78,6 +81,26 @@ func newFaissGPUFloat32Index(cpuIdx *faiss.IndexImpl, params *faissIndexParams)
 	return f, nil
 }
 
+func newFaissGPUFloat32IndexFromBytes(idxBytes []byte, params *faissIndexParams) (faissIndex, error) {
+	if idxBytes == nil {
+		return nil, errNilIndex
+	}
+
```
CascadingRadium, line 88: ditto - missing nil params check

faiss_vector_index_gpu_float32.go
```diff
@@ -130,6 +153,8 @@ func (f *faissGPUFloat32Index) close() {
 	f.waitGPU()
 	f.teardownGPU()
 	f.cpuIdx.Close()
+	f.cpuIdx = nil
```
CascadingRadium: i dont think the `f.cpuIdx = nil` is needed

CascadingRadium:
hey @Likith101, can you request @copilot for a review?

faiss_vector_index_gpu_float32.go
CascadingRadium: scratch that, I think this is fine. Can you add this index = nil in all the `Close()` APIs?

faiss_vector_index_bivf.go
Likith101: This is used only once the index is fully persisted in merge and query paths. So it should be readonly

centroid_index.go
CascadingRadium: here as well - process? We are certain that the process call is called when writing this data right?

CascadingRadium: need to process here as well like we do in the read path


```go
	// read the entierity of the docID list through the file reader
	buf, err := r.process(mem[pos : pos+int(listLen)])
	if err != nil {
		return nil, nil, nil, fmt.Errorf("could not process docID list: %v", err)
	}
	pos += int(listLen)
```

CascadingRadium: done

CascadingRadium (approved):
Pls add PR description before merge

capemox (approved):

PR #413, opened 2026-05-20
MB-71936: Set the correct ioflags for fastmerge

Likith101:
- FastMerge uses MergeFrom api within faiss which does not work with the ReadOnly io flag


capemox (approved):

CascadingRadium (approved):

CascadingRadium:
Please add PR description before merging and also please add a comment in the code for the decision to use the write flag.

PR #416, opened 2026-05-21
MB-71788: Use the training params for fast merge

Thejas-bhat:
- fast merge works by performing a training process prior to data ingestion and that means that the training parameters are controllable by the application layer such as cbft in couchbase. 
- so when a new vector index is being created as part of training, the config is provided with the training params that's used in the fast merge path

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -540,16 +540,8 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(trainedIndex faissIndexIV
 		return nil
 	}
 
-	// create the faiss index to hold the merged data, either via fast merge or reconstruction
-	nlist := determineCentroids(nvecs)
-	if trainedIndex == nil {
-		// when the training for fast merge is ongoing, we use the nlist value set in the
-		// config for the train process
-		if tp, ok := v.config[index.TrainingKey]; ok {
-			trainingParams := tp.(*index.TrainingParams)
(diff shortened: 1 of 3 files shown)


section_faiss_vector_index.go
```diff
@@ -814,11 +823,19 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *FileWriter) error {
...
 		}
 
+		// training key is associated with some addtional params that might be specific to
+		// the fast merge path
+		nlist := determineCentroids(nvecs)
+		if tp, ok := vo.config[index.TrainingKey]; ok {
+			trainingParams := tp.(*index.TrainingParams)
+			nlist = trainingParams.NumCentroids
+		}
+
```
CascadingRadium: add to a helper 

```go
func (vo *vectorIndexOpaque) numCentroids() int {
  	nlist := determineCentroids(nvecs)
	if tp, ok := vo.config[index.TrainingKey].(*index.TrainingParams); ok {
		nlist = tp.NumCentroids
	}
    return nlist
}
```

CascadingRadium (changes requested):

section_faiss_vector_index.go
Thejas-bhat: done

CascadingRadium (dismissed):

Likith101 (dismissed):

CascadingRadium (approved):

maneuvertomars (approved):

PR #417, opened 2026-05-22
MB-71954: Add gpu memory estimate

capemox:
Add new `estimateGPUMemSize` which is used to get a more accurate estimate of the memory occupied on GPU by an index.

### DIFF
--- a/faiss_vector_index_gpu_float32.go
+++ b/faiss_vector_index_gpu_float32.go
@@ -140,13 +140,64 @@ func (f *faissGPUFloat32Index) initGPU() {
 	if !f.canUseGPU() {
 		return
 	}
-	gpuIdx, err := faiss.CloneToGPU(f.cpuIdx)
+	gpuIdx, err := faiss.CloneToGPU(f.cpuIdx, f.estimateGPUMemSize())
 	if err != nil || gpuIdx == nil {
 		return
 	}
 	gs := &gpuState{idx: gpuIdx}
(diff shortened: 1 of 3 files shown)


CascadingRadium:
hey @capemox, i think the wrong api is used, we cannot use the `ObtainClusterVectorCountsFromIVFIndex` API as it crashes due to the assert in FAISS - `FAISS_ASSERT(params != nullptr && params->sel != nullptr);`. 

The API was designed with filtering in mind and does not work with no selector. 

Moreover, i think i found a better method (sorry for not pointing it out earlier, i just found it) - There is an API `getGpuVectorsEncodingSize_` in `IVFFlat.h` and it does this exact size estimation for interleaved / legacy layout. 

Since we already have the mapping of list -> nvecs in that list, (you can add a custom API for this if you want (which does not take in a selector) - just dont use the cluster vector count API). We need to use this `getGpuVectorsEncodingSize_` per list as we pad list-wise on FAISS side (i suppose).

```python
func getMemEstimate:
   vl = []int of size nlist (vl[i] = 10 => 10 vecs in ith list) 
   vl = getVl()
   totSize = 0
   for listID, nvecs in vl:
       totSize += getGpuVectorsEncodingSize_(nvecs)
   return totSize
```

You can add a C binding for the API and perhaps shift this logic to go-faiss instead, as we would need to keep this stuff under the `gpu` tag.

Thanks for the help~!

PR #420, opened 2026-05-28
MB-72240: use the vector cache for fast merge

Thejas-bhat:
- keep the trained index in the trained index cache to reduce the garbage
generated during fast merge.
  - the index is always kept in the cache so that we don't prematurely close the index entry especially when the merge operations are sporadic
- fast merge requires only the vector index, so we skip constructing the
id mapping


centroid_index.go
```diff
@@ -48,37 +48,9 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
...
-	pos += uint64(n)
-	pos += mapLen
-
-	// type of index
-	indexType, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-	pos += uint64(n)
-	indexSize, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-	pos += uint64(n)
+	useGPU := sb.fieldsOptions[field].UseGPU()
+	index, _, _, err := sb.vecIndexCache.loadOrCreate(fieldIDPlus1-1, sb.mem[pos:],
```
CascadingRadium: since we only need the index here, can we not have another API in the cache that takes in the required argument only and avoid the overhead of creating/returning the idmapping and the bitmap?

Thejas-bhat: yeah that's right, i'm still implementing that bit

faiss_vector_cache.go
```diff
@@ -62,9 +62,21 @@ func (vc *vectorIndexCache) Clear() {
...
+	optStr  string
+	// while performing fast merge we skip the construction of the id mapping since its
+	// not relevant to that flow
+	skipMapping bool
+}
+
 // loadOrCreate obtains the vector index from the cache or creates it if it's not present.
-// useGPU indicates whether the field mapping requires GPU acceleration for this index.
-func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, numDocs uint32, except *roaring.Bitmap, useGPU bool, r *FileReader, optStr string) (
+func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, opts loadOrCreateOptions) (
```
capemox: opts could be a pointer rather than a copy

faiss_vector_cache.go
```diff
@@ -89,18 +101,17 @@ func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, numDocs uin
...
 	}
 	// still not present, create and cache it
-	return vc.createAndCacheLOCKED(fieldID, mem, numDocs, except, useGPU, r, optStr)
+	return vc.createAndCacheLOCKED(fieldID, opts)
 }
 
-// Rebuilding the cache on a miss.
-func (vc *vectorIndexCache) createAndCacheLOCKED(fieldID uint16, mem []byte,
-	numDocs uint32, except *roaring.Bitmap, useGPU bool, r *FileReader, optStr string) (index faissIndex,
+func (vc *vectorIndexCache) createAndCacheLOCKED(fieldID uint16, opts loadOrCreateOptions) (index faissIndex,
```
capemox: ditto

section_faiss_vector_index.go
```diff
@@ -249,13 +250,13 @@ func (v *faissVectorIndexSection) Merge(opaque map[int]resetable, segments []*Se
 	return nil
 }
 
-func trainedIndexFromConfig(config map[string]interface{}, fieldName string) (faissIndexIVF, error) {
+func trainedIndexFromConfig(config map[string]interface{}, fieldName string) (faissIndexIVF, func() error, error) {
 	if config == nil {
-		return nil, nil
+		return nil, nil, nil
```
capemox: returned callback should be `func() error { return nil }` like later in the function

capemox (changes requested):

trained_index.go
```diff
@@ -48,37 +69,21 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
...
-
-	// length of the vector to docID map
-	mapLen, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-	pos += uint64(n)
-	pos += mapLen
+	useGPU := sb.fieldsOptions[field].UseGPU()
+	index, _, _, err := sb.vecIndexCache.loadOrCreate(fieldIDPlus1-1, loadOrCreateOptions{
+		mem:         sb.mem[pos:],
+		numDocs:     uint32(sb.numDocs),
+		useGPU:      useGPU,
```
Likith101: I don't think putting the centroid index into the gpu is required? Won't it add more downtime to transfer data to and from the GPU? I think we should be passing in a false here directly

Thejas-bhat: yeah makes sense, thanks

section_faiss_vector_index.go
Thejas-bhat: yeah thanks, i've substituted for all the return statements

faiss_vector_cache.go
Thejas-bhat: i thought it was okay to have it as a struct over here, since it'll reside on the stack (so its not going to add pressure to the GC) and each search thread anyways has to create a struct over here. the copy would occur only for the thread that's actually going to create the index and register into cache. this should be an acceptable tradeoff, since way we won't have to add a lot of nil checks in the code. please let me know if that makes sense

Likith101 (dismissed):

capemox (dismissed):

CascadingRadium:
Here is my opinion, From what I can gather, we load the trained index into the cache the first time we hit the API and then reuse it for every merge operation to enable fast merge.

My concern with this approach is that the vector index cache is designed for the query path, with the EWMA mechanism keeping the index alive for a short window after its reference count drops to zero. 

Reusing this same struct for indexing is not ideal. We can easily end up evicting the trained index the moment `closeTrainedIndex()` is called. 

Since we `incRef` and `decRef` the index within a single merge cycle, and merges are both infrequent and happen at highly variable intervals, we can easily end up thrashing the cache.

Additionally, modifying a cache that was purpose-built for the query path introduces fragility, with fast merge logic interfering into query path code, making it harder to reason about and more prone to breakage as new features are added.

My proposal: add a dedicated always-alive cache in trained_index.go. This would be a new type that holds only the trained index and whatever state is required exclusively for the fast merge path. The EWMA eviction mechanism would not apply to it, the trained index would remain available for the lifetime of the parent index, similar to the _nested index cache_, which permanently persists the ancestry chain. This way throughout the index lifetime, irrespective of when a merge occurred, we always have the trained index at hand.

Since we memory-map the faiss index anyway, we are not losing out on performance there. This always-alive cache will be applicable only for the `trained segment`, and the regular cache will be applicable for the regular segments. 


```go
func newTrainedIndexCache() *trainedIndexCache {
	return &trainedIndexCache{
		cache: make(map[uint16]faissIndex),
	}
}

type trainedIndexCache struct {
	m     sync.RWMutex
	cache map[uint16]faissIndex
        /// No EWMA or eviction
}

// loadOrCreate obtains the trained faiss index from the cache or creates it
// from the provided segment memory if not already present.
func (tc *trainedIndexCache) loadOrCreate(fieldID uint16) (faissIndex, error) {
	tc.m.RLock()
	if idx, ok := tc.cache[fieldID]; ok {
		tc.m.RUnlock()
		return idx, nil
	}
	tc.m.RUnlock()

	tc.m.Lock()
	defer tc.m.Unlock()

	if idx, ok := tc.cache[fieldID]; ok {
		return idx, nil
	}

	idx, err := tc.createFromOptsLOCKED(opts)
	if err != nil {
		return nil, err
	}
	tc.cache[fieldID] = idx
	return idx, nil
}

// Clear closes all cached trained indices and clears the cache.
func (tc *trainedIndexCache) Clear() {
	tc.m.Lock()
	defer tc.m.Unlock()
	for _, idx := range tc.cache {
		idx.close()
	}
	tc.cache = nil
}

// call clear when we close the segment, along with the other cache clear ops.
	s.nstIndexCache.Clear()
	s.trainedIndexCache.Clear()
```

CascadingRadium (changes requested):

Thejas-bhat:
> modifying a cache that was purpose-built for the query path introduces fragility, with fast merge logic interfering into query path code, making it harder to reason about and more prone to breakage as new features are added.

yeah makes sense, thanks. i'll make the changes

faiss_vector_cache.go
```diff
@@ -89,106 +100,117 @@ func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, numDocs uin
...
 	pos += n
 
 	// read the entierity of the docID list through the file reader
-	buf, err := r.process(mem[pos : pos+int(listLen)])
+	buf, err := opts.reader.process(mem[pos : pos+int(listLen)])
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("could not process docID list: %v", err)
+		return nil, nil, fmt.Errorf("could not process docID list: %v", err)
 	}
 	pos += int(listLen)
```
CascadingRadium, line 152: should this not be within the `if !opts.skipMapping {` block?

```go
listPos := pos
pos += int(listLen)

if !opts.skipMapping {
	// read the entierity of the docID list through the file reader
	buf, err := opts.reader.process(mem[listPos : listPos+int(listLen)])
	if err != nil {
		return nil, nil, fmt.Errorf("could not process docID list: %v", err)
	}
... // read idMapping from buf

}
```

segment.go
```diff
@@ -183,6 +185,7 @@ func (sb *SegmentBase) DecRef() (err error) { return nil }
 func (sb *SegmentBase) Close() (err error) {
 	sb.invIndexCache.Clear()
 	sb.vecIndexCache.Clear()
+	sb.trainedIndexCache.Clear()
```
CascadingRadium, line 188: need to `Clear` it in the below method as well
```
func (s *Segment) closeActual() (err error) {
```

trained_index.go
```diff
@@ -48,37 +48,14 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
...
-		binaryIndexSize, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-		pos += uint64(n)
-		bIndexBytes, err := sb.fileReader.process(sb.mem[pos : pos+binaryIndexSize])
-		if err != nil {
-			return nil, err
-		}
-		return newFaissBinaryIndexFromBytes(bIndexBytes, fIndexBytes, params)
-	}
-	return newFaissFloat32IndexFromBytes(fIndexBytes, params)
+	index, err := sb.trainedIndexCache.loadOrCreate(fieldIDPlus1-1, loadOrCreateOptions{
```
CascadingRadium: can just return here.

faiss_vector_cache.go
```diff
@@ -62,9 +62,20 @@ func (vc *vectorIndexCache) Clear() {
 	vc.m.Unlock()
 }
 
+// loadOrCreateOptions controls what loadOrCreate builds and returns.
+type loadOrCreateOptions struct {
```
CascadingRadium: although this was already discussed before, i feel like we should have a constructor for this as `newVectorCacheOptions` that returns a pointer to the struct. Right now we create 2 copies of the struct, when creating the cache entry (readVectorSection) and 1 copy when loading the cached index.

Thejas-bhat: since both of you have suggested that its acceptable to have it as a pointer instead, i've included this in the patch

Thejas-bhat (commented):
yeah my bad, i forgot to push the patch that was there on my system

faiss_vector_cache.go
```diff
@@ -89,106 +116,122 @@ func (vc *vectorIndexCache) loadOrCreate(fieldID uint16, mem []byte, numDocs uin
...
-	// create a mapping using the numVecs and numDocs
-	mapping = newIDMapping(uint32(numVecs), numDocs)
-	for vecID := uint32(0); vecID < uint32(numVecs); vecID++ {
-		docID, n := binary.Uvarint(buf[bufPos:min(bufPos+binary.MaxVarintLen64, bufLen)])
-		if n <= 0 {
-			return nil, nil, nil, fmt.Errorf("could not read docID for vecID %d", vecID)
+
+	if !opts.skipMapping {
+		// read the entierity of the docID list through the file reader
+		buf, err := opts.reader.process(mem[pos : pos+int(listLen)])
```
CascadingRadium: this is a bug, pos was bumped by listLen in L151 but we are reading from the ptr which is ahead of where it should be.

buf = (pos:pos+listLen).

copy the old pos value in L150 into a local var like `listPos` and use that here.

trained_index.go
```diff
@@ -48,37 +48,6 @@ func (sb *SegmentBase) GetCoarseQuantizer(field string) (interface{}, error) {
...
-		binaryIndexSize, n := binary.Uvarint(sb.mem[pos : pos+binary.MaxVarintLen64])
-		pos += uint64(n)
-		bIndexBytes, err := sb.fileReader.process(sb.mem[pos : pos+binaryIndexSize])
-		if err != nil {
-			return nil, err
-		}
-		return newFaissBinaryIndexFromBytes(bIndexBytes, fIndexBytes, params)
-	}
-	return newFaissFloat32IndexFromBytes(fIndexBytes, params)
+	opts := newVectorCacheOptions(sb.mem, uint32(sb.numDocs), nil, false, sb.fileReader, optStr, true)
```
CascadingRadium: sb.mem[pos:]?

faiss_vector_posting.go
```diff
@@ -304,8 +304,10 @@ func (sb *SegmentBase) InterpretVectorIndex(field string, except *roaring.Bitmap
 	// create the vector index wrapper by loading (or creating) the vector index
 	// and the vector to docID mapping
 	useGPU := sb.fieldsOptions[field].UseGPU()
+
 	var err error
-	rv.index, rv.mapping, rv.exclude, err = sb.vecIndexCache.loadOrCreate(fieldID, sb.mem[pos:], uint32(sb.numDocs), except, useGPU, sb.fileReader, optStr)
+	opts := newVectorCacheOptions(sb.mem, uint32(sb.numDocs), except, useGPU, sb.fileReader, optStr, false)
```
CascadingRadium: sb.mem[pos:]

faiss_vector_cache.go
Thejas-bhat: yeah my bad, thanks

CascadingRadium (approved):

Likith101 (approved):

PR #422, opened 2026-06-01
Bug Fix: Add a guard for config

CascadingRadium:
- `config` maybe `nil` when using the legacy `New` and `Open` APIs.
- Add a guard to always fallback to naive merge if `config` is nil in the fastmerge codepath.

### DIFF
--- a/section_faiss_vector_index.go
+++ b/section_faiss_vector_index.go
@@ -722,7 +722,7 @@ func determineCentroids(nvecs int) int {
 func (vo *vectorIndexOpaque) numCentroids(nvecs int) int {
 	nlist := determineCentroids(nvecs)
 	if vo.config != nil {
-		// training key is associated with some addtional params such as num centroids
+		// training key is associated with some additional parameters such as num centroids
 		// that might be specific to the fast merge path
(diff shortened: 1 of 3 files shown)


capemox (approved):

section_faiss_vector_index.go
```diff
@@ -250,6 +250,9 @@ func (v *faissVectorIndexSection) Merge(opaque map[int]resetable, segments []*Se
 }
 
 func trainedIndexFromConfig(config map[string]interface{}, fieldName string) (faissIndexIVF, error) {
+	if config == nil {
+		return nil, nil
```
Samsonnyyeet, line 254: Should we return an error here? Maybe mentioning that config is nil.

CascadingRadium: no -> config being nil here indicates that fast merge is not supported, which means that we fallback to using naive merge. Returning an error in the merge path here would cause the process to crash.

Samsonnyyeet: Cool!

Samsonnyyeet (approved):

PR #432, opened 2026-06-26
MB-72376: Introducing New Geo-Shape Indexes

Likith101:
- Introducing a new geo shape index section and section specific cache
- Added methods to get the geoData from a segment
- Separate ewma from vector cache to be reused within geo cache


section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+	}
+
+	// Append cross cells along with their corresponding doc IDs
+	g.crossCells = append(g.crossCells, crossCells...)
+	for _ = range crossCells {
+		g.crossDocIDs = append(g.crossDocIDs, docID)
+	}
+
+	// Append bounding box bytes, shape bytes and the document score
+	g.boundingBoxes = append(g.boundingBoxes, f.EncodedBoundingBox())
```
Thejas-bhat, line 301: since these structures are called once for every doc, can you prealloc the slice capacity to numDocs?

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+	g.shapes = append(g.shapes, f.EncodedShape())
+	innerScore, crossScore := f.Scores()
+	g.docScoresInner = append(g.docScoresInner, innerScore)
+	g.docScoresCross = append(g.docScoresCross, crossScore)
+}
+
+func (g *geoShapeV2IndexSectionOpaque) persist(w *FileWriter) error {
+
```
Thejas-bhat: nit: can you drop these kind of empty lines?

CascadingRadium (dismissed):

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+	// newDocNums maps a document's old segment doc number (the index) to its
+	// doc number in the merged segment, with docDropped marking deleted docs.
+	newDocNums []uint64
+}
+
+func (g *geoShapeV2IndexSection) Merge(opaque map[int]resetable, segments []*SegmentBase,
+	drops []*roaring.Bitmap, fieldsInv []string, newDocNumsIn [][]uint64, w *FileWriter,
+	closeCh chan struct{}) error {
+
```
Thejas-bhat: nit: remove new line

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+		crossCells:     make([]uint64, 0, totalCross),
+		crossDocIDs:    make([]uint32, 0, totalCross),
+		boundingBoxes:  make([][]byte, 0, totalDocs),
+		shapes:         make([][]byte, 0, totalDocs),
+	}
+
+	// Assign merged geo docIDs and build, per segment, a direct
+	// oldGeoDocID -> mergedGeoDocID slice
+	segRemaps, numDocs := buildGeoDocRemaps(indexInfos, mergedContent)
+
```
Thejas-bhat: nit: maybe you can remove these kind of new lines?

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+	if numDocs == 0 {
+		return nil, nil
+	}
+
+	// Each segment's inner and cross cells are already stored sorted
+	// So instead of concatenating every segment's cells and sorting the whole
+	// set, we k-way merge the pre-sorted per-segment runs while.
```
Thejas-bhat, line 528: does the comment need to be updated?

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+		docScoresCross: make([]uint64, 0, totalDocs),
+		innerCells:     make([]uint64, 0, totalInner),
+		innerDocIDs:    make([]uint32, 0, totalInner),
+		crossCells:     make([]uint64, 0, totalCross),
+		crossDocIDs:    make([]uint32, 0, totalCross),
+		boundingBoxes:  make([][]byte, 0, totalDocs),
+		shapes:         make([][]byte, 0, totalDocs),
+	}
+
+	// Assign merged geo docIDs and build, per segment, a direct
```
Thejas-bhat, line 519: just to clarify - geoDocIDs are just the docIDs of the docs having geo_v2 field content right?

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+	for s, indexInfo := range indexInfos {
+		remap := make([]uint32, len(indexInfo.content.docNums))
+		for geoDocID, oldDocNum := range indexInfo.content.docNums {
+			newDocNum := indexInfo.newDocNums[oldDocNum]
+			if newDocNum == docDropped {
+				remap[geoDocID] = uint32(math.MaxUint32)
+				continue
+			}
+			remap[geoDocID] = uint32(numDocs)
+			numDocs++
```
Thejas-bhat: since numDocs corresponds to the geoDocIDs, can you rename this to numGeoDocs?

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+	if !g.init {
+		g.alloc()
+		g.init = true
+	}
+
+	innerCells := f.InnerCells()
+	crossCells := f.CrossCells()
+
+	docID := uint32(len(g.docNums))
+	g.docNums = append(g.docNums, docNum)
```
Thejas-bhat, line 286: i'm guessing this is a mapping of geoDocIDs -> localDocNum (of the one having this geoshape field) right? can you rename docID to geoDocID? its a little confusing otherwise 
and also going by that assumption, if one of the local docs doesn't have the geo shape field in it, then the mapping and the geoDocID generation would be such that we don't faulty mappings right? and you've verified in your functional testing?

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,737 @@
...
+
+	return mergedContent, nil
+}
+
+// buildGeoDocRemaps assigns the merged segment's geo docIDs and returns, for
+// each input segment, a slice mapping that segment's old geo docID to its
+// merged geo docID.
+func buildGeoDocRemaps(indexInfos []*geoIndexInfo,
+	mergedContent *geoIndexContent) (segRemaps [][]uint32, numDocs uint64) {
+
```
Thejas-bhat: nit: remove the empty line

section_geo_shape_v2_index.go
Thejas-bhat: i think there are some funcs that have these kind of empty new lines. so can you please keep them consistent in all funcs?

Thejas-bhat (changes requested):

CascadingRadium (dismissed):

geo_index_cache.go
```diff
@@ -0,0 +1,465 @@
...
+}
+
+func (gc *geoIndexCache) loadOrCreate(field uint16, mem []byte, except *roaring.Bitmap, r *FileReader) (*geoData, error) {
+	gc.m.RLock()
+	if gc.isClosed {
+		gc.m.RUnlock()
+		return nil, nil
+	}
+
+	entry, ok := gc.cache[field]
```
maneuvertomars, line 68: Is there a race case here?
we release the read lock before calling entry.load(). In that window cleanup() can take the write lock, see refs == 0, and evict + Close() this entry, after which we incRef() an orphaned entry. It's harmless today since Close() only decrements a refcount, but it becomes a use(only it gets FREE) the moment Close() frees real resources (the case the faiss vector cache guards against).

this might help:


```
if entry, ok := gc.cache[field]; ok {
    defer gc.m.RUnlock()
    return entry.load(except), nil
}
```

geo_index_cache.go
```diff
@@ -0,0 +1,465 @@
...
+
+// createNewExcludeBitmap translates an exclusion bitmap from segment doc
+// number space into geo docID space
+func createNewExcludeBitmap(except *roaring.Bitmap, docNums []uint32) *roaring.Bitmap {
+	if except == nil || except.IsEmpty() {
+		return nil
+	}
+
+	newExcept := roaring.New()
+	for i, docNum := range docNums {
```
maneuvertomars: This scans all of docNums on every GeoShapeV2Data call. Since docNums is sorted ascending, iterating the (typically much smaller) except bitmap and binary searching docNums would make this O(|except|·log n) per query instead of O(n).

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,733 @@
...
+	innerCells := f.InnerCells()
+	crossCells := f.CrossCells()
+
+	geoDocID := uint32(len(g.docNums))
+	g.docNums = append(g.docNums, docNum)
+
+	// Append inner cells along with their corresponding doc IDs
+	g.innerCells = append(g.innerCells, innerCells...)
+	for _ = range innerCells {
```
maneuvertomars: we can change it to 
`for range innerCells`

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,733 @@
...
+		g.innerDocIDs = append(g.innerDocIDs, geoDocID)
+	}
+
+	// Append cross cells along with their corresponding doc IDs
+	g.crossCells = append(g.crossCells, crossCells...)
+	for _ = range crossCells {
```
maneuvertomars: same as above.

file_callbacks.go
```diff
@@ -97,6 +114,151 @@ func (w *FileWriter) Sum32() uint32 {
...
+func (w *FileWriter) grabPayloadBuf(size int) []byte {
+	if cap(w.payloadBuf) < size {
+		w.payloadBuf = make([]byte, size)
+	}
+	return w.payloadBuf[:size]
+}
+
+// WriteUint64Array writes arr as a length-prefixed array of little-endian
+// uint64 values, padded so the payload begins on an 8-byte boundary in the
+// file. Padding is required for zero-copy reads becaus of go's
```
maneuvertomars: becaus -> because.

section_geo_shape_v2_index.go
```diff
@@ -0,0 +1,733 @@
...
+func (g *geoShapeV2IndexSection) getGeoShapeV2IndexOpaque(
+	opaque map[int]resetable) *geoShapeV2IndexSectionOpaque {
+	if _, ok := opaque[SectionGeoShapeV2Index]; !ok {
+		opaque[SectionGeoShapeV2Index] = g.InitOpaque(nil)
+	}
+	return opaque[SectionGeoShapeV2Index].(*geoShapeV2IndexSectionOpaque)
+}
+
+type geoShapeV2IndexSectionOpaque struct {
+	results []index.Document
```
maneuvertomars: i guess this is being written, but not being read anywhere. pl verify & remove if it is so.

maneuvertomars (changes requested):

Thejas-bhat (approved):

maneuvertomars (approved):

PR #435, opened 2026-07-13
[WIP] perf: cache FST term-offset lookups in Dictionary.postingsList

capemox:
Cache FST term-offset lookups so repeated queries on hot terms skip re-traversing the FST. No functional/API changes; benchmark included.

### Changes

- **Cache FST-resolved term offset** — `Dictionary.postingsList()` resolves each term through the FST on every call, even though the offset never changes for a given (segment, field, term). Cache it in a per-field write-once/read-many `sync.Map` on the segment's inverted-index cache entry; warm lookups jump straight to `postingsListFromOffset`. `BenchmarkPostingsListLookup` (fully-warm): **243ns → 75ns (-69%)**, allocations unchanged.
- **Cache FST not-found results** — extend the same cache to record misses via a `termNotFoundSentinel`, so high-selectivity / rare-entity disjunction queries stop re-traversing the FST for absent terms.

Real-world gain scales with term repetition across queries (the first lookup per term still pays the traversal). Marked draft pending review.

### DIFF
--- a/inverted_text_cache.go
+++ b/inverted_text_cache.go
@@ -127,7 +127,8 @@ type invertedCacheEntry struct {
 	// termOffsetCache maps term → posting-list offset within the segment
 	// file, avoiding repeated FST traversals for repeated queries on the
 	// same term.  Uses sync.Map (write-once, read-many pattern).
-	// Only populated for terms that are found in the segment.
+	// Populated for both found terms (the resolved offset) and absent terms
+	// (termNotFoundSentinel), so repeated misses also skip the FST.
 	termOffsetCache sync.Map // key: string, val: uint64
 }
 
--- a/inverted_text_cache.go
+++ b/inverted_text_cache.go
@@ -97,9 +97,38 @@ func (sc *invertedIndexCache) insertLOCKED(fieldID uint16, fst *vellum.FST) {
 	}
 }
 
+// getOrCreateEntry returns the invertedCacheEntry for fieldID, creating an
+// empty entry if none exists yet (for segments that have no FST loaded but
+// still need the termOffset cache).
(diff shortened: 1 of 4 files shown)


inverted_text_cache.go
```diff
@@ -97,9 +97,39 @@ func (sc *invertedIndexCache) insertLOCKED(fieldID uint16, fst *vellum.FST) {
...
 // invertedCacheEntry is the vellum FST and is the value stored in the invertedIndexCache cache, for a given fieldID.
 type invertedCacheEntry struct {
 	fst *vellum.FST
+
+	// termOffsetCache maps term → posting-list offset within the segment
+	// file, avoiding repeated FST traversals for repeated queries on the
+	// same term.  Uses sync.Map (write-once, read-many pattern).
+	// Populated for both found terms (the resolved offset) and absent terms
+	// (termNotFoundSentinel), so repeated misses also skip the FST.
+	termOffsetCache sync.Map // key: string, val: uint64
```
Thejas-bhat, line 132: wouldn't this cause memory issues though? in a real world scenario where the segments are big containing large FSTs, this map could explode in size when most of the entries in the FSTs are queried. 
since there's no prefix sharing and this is purely present in the memory, storage will be amplified along with the extra copy of the term in FST. i don't think this field would be compatible with mmap either, since it doesn't demand load the specific term from the mmap'd slice like vellum. 

also, another question, when you performed the benchmarking what was the dataset used and how many documents where there in the biggest segment?

capemox: I was using a relatively small benchmark so nothing problematic showed up. You're right, this is a problem. Anyways I'm going to pull the fuzzy stuff into a separate PR, I'll check it again.

CascadingRadium, line 132: also i dont see any eviction mechanism here?

PR #436, opened 2026-07-17
MB-72489: omit postings-count read during automaton candidate collection

capemox:
Split out from #435 per review — the automaton (fuzzy/regexp) candidate-collection optimization, as its own PR. No functional/API changes.

### Change

`DictionaryIterator.Next()` unconditionally deserialized each visited term's postings list (roaring `FromBuffer` + `GetCardinality`) to populate `DictEntry.Count`. Candidate-term collectors in bleve (fuzzy/regexp/wildcard) discard that count, so it was O(candidates) wasted bitmap deserializations per query. The `omitCount` field existed to suppress this but was never wired up.

New `AutomatonIteratorOmitCount` returns an iterator that leaves `Count` at zero and skips the postings read; the existing `AutomatonIterator` is unchanged (both share a helper). Consumed opt-in by bleve's fuzzy/regexp field-dict constructors (blevesearch/bleve#2382).

### Benchmark

`BenchmarkFuzzyIterator` (1000-term / 3000-doc segment, fuzziness-2 automaton matching 280 multi-hit candidates), Apple M4 Pro, count=10 via benchstat, with-count vs omit-count:

| sec/op | B/op | allocs/op |
|---|---|---|
| **−35.5%** | **−57.4%** | **−47.7%** (595→311) |

All p=0.000, n=10. `TestAutomatonIteratorOmitCountSemantics` (added) asserts the omit iterator yields identical terms/order/edit-distances — only the discarded count differs.

### DIFF
--- a/dict.go
+++ b/dict.go
@@ -119,9 +119,25 @@ func (d *Dictionary) Contains(key []byte) (bool, error) {
 // having the the vellum automaton and start/end key range
 func (d *Dictionary) AutomatonIterator(a segment.Automaton,
 	startKeyInclusive, endKeyExclusive []byte) segment.DictionaryIterator {
+	return d.automatonIterator(a, startKeyInclusive, endKeyExclusive, false)
+}
+
+// AutomatonIteratorOmitCount is like AutomatonIterator, but the returned
+// iterator does not populate DictEntry.Count. Skipping the count avoids a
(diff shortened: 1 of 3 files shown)


steveyen (approved):
lgtm

Thejas-bhat (approved):

PR #441, opened 2026-08-11
MB-72082: add more stats around vector search

Thejas-bhat:
- introduce a new stats handler in the zap plugin. this would benefit in tracking stats all over the codebase using a single *Stats instance in the plugin object. 
- currently this PR covers stats for the vector section on float indexes, but the handler can be passed around with new stats to other sections as well to get a better deep dive view of the zap layer.
- cleanup a vector unit test


capemox (dismissed):

merge.go
```diff
@@ -48,7 +49,7 @@ func (z *ZapPlugin) MergeUsing(segments []seg.Segment, drops []*roaring.Bitmap,
 	return z.merge(segments, drops, path, closeCh, s, config)
 }
 
-func (*ZapPlugin) merge(segments []seg.Segment, drops []*roaring.Bitmap, path string,
+func (z *ZapPlugin) merge(segments []seg.Segment, drops []*roaring.Bitmap, path string,
```
Likith101: this changed line is no longer necessary right?

merge.go
```diff
@@ -62,11 +63,36 @@ func (*ZapPlugin) merge(segments []seg.Segment, drops []*roaring.Bitmap, path st
 			panic(fmt.Sprintf("oops, unexpected segment type: %T", segment))
 		}
 	}
-	return mergeSegmentBases(segmentBases, drops, path, DefaultChunkMode, closeCh, s, config)
+
+	zapStats, ok := config[seg.StatsKey].(*seg.Stats)
```
maneuvertomars, line 67: A typed nil *seg.Stats stored under StatsKey passes this `ok` check and panics on the first atomic add. Can we make it `if !ok || zapStats == nil?`(below line 68)

segment.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/RoaringBitmap/roaring/v2"
 	index "github.com/blevesearch/bleve_index_api"
 	mmap "github.com/blevesearch/mmap-go"
+	seg "github.com/blevesearch/scorch_segment_api/v2"
```
maneuvertomars: we can drop this duplicate import.

section_inverted_text_index.go
```diff
@@ -1045,6 +1045,8 @@ type invertedIndexOpaque struct {
 
 	fieldsSame bool
 	numDocs    uint64
+
+	stats *seg.Stats
```
maneuvertomars: is it ever assigned or read? there is no mention of "stats" in Set(). we can remove or wire it.

section_faiss_vector_index.go
```diff
@@ -852,10 +891,19 @@ func (vo *vectorIndexOpaque) writeVectorIndexes(w *FileWriter) error {
...
 }
 
 func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, docNum uint32) {
+	start := time.Now()
+	defer func() {
+		atomic.AddUint64(&vo.stats.TotVecSectionVecsProcessedTime, uint64(time.Since(start)))
+		if vo.trainingPhase {
+			atomic.AddUint64(&vo.stats.TotVecSectionTrainingPhaseVecsProcessedTime, uint64(time.Since(start)))
+		}
+	}()
```
maneuvertomars, line 906: time.Since(start) is called twice, so the training phase counter records a slightly larger duration than the total it's a subset of, the two can never reconcile. we can compute it once and can use something like:
```
defer func() {
    elapsed := uint64(time.Since(start))
    atomic.AddUint64(&vo.stats.TotVecSectionVecsProcessedTime, elapsed)
    if vo.trainingPhase {
        atomic.AddUint64(&vo.stats.TotVecSectionTrainingPhaseVecsProcessedTime, elapsed)
    }
}()
```

maneuvertomars (changes requested):

merge.go
Thejas-bhat: thanks, i've addressed all your review comments

new.go
```diff
@@ -50,10 +50,16 @@ func (z *ZapPlugin) NewUsing(results []index.Document, config map[string]interfa
 	return z.newWithChunkMode(results, DefaultChunkMode, config)
 }
 
-func (*ZapPlugin) newWithChunkMode(results []index.Document,
+func (z *ZapPlugin) newWithChunkMode(results []index.Document,
 	chunkMode uint32, config map[string]interface{}) (segment.Segment, uint64, error) {
 	s := interimPool.Get().(*interim)
 
+	zapStats, ok := config[segment.StatsKey].(*segment.Stats)
+	if !ok {
```
capemox: Add a `zapStats == nil` check to be safe.

capemox (changes requested):

maneuvertomars (dismissed):

capemox (approved):

Likith101 (approved):

section_faiss_vector_index.go
```diff
@@ -883,8 +933,8 @@ func (vo *vectorIndexOpaque) process(field index.VectorField, fieldID uint16, do
 				dimension:    dim,
 				metric:       metric,
 				optimizedFor: indexOptimizedFor,
-				vectors:      make([]float32, 0, dim*numVectors),
-				vecDocIDs:    make([]uint32, 0, numVectors),
+				vectors:      make([]float32, 0, dim*numVectors*vo.numDocs),
+				vecDocIDs:    make([]uint32, 0, numVectors*vo.numDocs),
```
maneuvertomars, line 937: This prealloc is dim * numVectors * numDocs, but numVectors comes from the first doc only. A multi vector first doc in a large batch reserves GBs up front, and a big enough product overflows and panics in makeslice. Can we go back to dim*numVectors and let append grow it?

section_faiss_vector_index.go
```diff
@@ -978,6 +1038,16 @@ func (v *vectorIndexOpaque) Set(key string, val interface{}) {
 		v.fieldsOptions = val.(map[string]index.FieldIndexingOptions)
 	case "config":
 		v.config = val.(map[string]interface{})
+		if v.config != nil {
+			if tp, ok := v.config[index.TrainingKey].(*index.TrainingParams); ok && tp != nil {
+				v.trainingPhase = true
```
maneuvertomars, line 1043: trainingPhase is only ever set to true, and Reset() doesn't clear it. Since interims are pooled, one training build flips it forever and later normal builds keep inflating the training phase counters. We can do like v.trainingPhase = (tp != nil) in Set("config") and clearing trainingPhase/numDocs/stats in Reset().

section_faiss_vector_index.go
```diff
@@ -212,6 +215,8 @@ func (v *faissVectorIndexSection) Merge(opaque map[int]resetable, segments []*Se
 				continue
 			}
 
+			atomic.AddUint64(&vo.stats.TotVecSectionVecsDeleted, uint64(newIndexInfo.nvecs-len(newIndexInfo.vecIds)))
```
maneuvertomars, line 218: This sits after the len(vecIds) == 0 continue at line 213, so segments whose vectors were all deleted add nothing to the deleted count. We can move it above the guard.

section_faiss_vector_index.go
```diff
@@ -251,6 +259,10 @@ func (v *faissVectorIndexSection) Merge(opaque map[int]resetable, segments []*Se
 			return err
 		}
 	}
+
+	if totalVecFields > int(atomic.LoadUint64(&vo.stats.TotVecSectionFieldsIndexed)) {
+		atomic.StoreUint64(&vo.stats.TotVecSectionFieldsIndexed, uint64(totalVecFields))
```
maneuvertomars, line 264: This load-then-store max isn't atomic, two concurrent merges can interleave and the smaller value wins. Also the Tot prefix suggests a counter, not a max, and the flush path never updates it. Either we can do  atomic.AddUint64, or a CAS loop with a Max* name.

merge.go
```diff
@@ -62,11 +63,36 @@ func (*ZapPlugin) merge(segments []seg.Segment, drops []*roaring.Bitmap, path st
...
+	atomic.AddUint64(&zapStats.TotMergeInputSegments, uint64(len(segments)))
+	var totalInputDocs, droppedDocs uint64
+	for i, sb := range segmentBases {
+		totalInputDocs += sb.numDocs
+		if drops[i] != nil {
+			droppedDocs += drops[i].GetCardinality()
+		}
+	}
+	atomic.AddUint64(&zapStats.TotMergeDroppedDocs, droppedDocs)
+	atomic.AddUint64(&zapStats.TotMergeOutputDocs, totalInputDocs-droppedDocs)
```
maneuvertomars, line 82: These are incremented before the merge runs, so aborted/failed merges still count their docs and retries double count. TotMergeOutputDocs is also input-minus-drops, not actual output. Can we count these after success, using computeNewDocCount's result?

build.go
```diff
@@ -99,7 +108,7 @@ func rewriteSegmentBase(sb *SegmentBase, path string) error {
 	closeCh := make(chan struct{})
 	defer close(closeCh)
 	_, _, err := mergeSegmentBases([]*SegmentBase{sb}, []*roaring.Bitmap{nil},
-		path, DefaultChunkMode, closeCh, nil, nil)
+		path, DefaultChunkMode, closeCh, nil, nil, sb.stats)
```
maneuvertomars, line 111: rewriteSegmentBase passes sb.stats into the merge machinery, so a persist (callback rewrite) inflates the vector-section merge counters while TotMerges* stays untouched. Should this path skip stats, or be counted as a persist?

section_faiss_vector_index.go
```diff
@@ -469,6 +485,14 @@ func (v *vectorIndexOpaque) mergeAndWriteVectorIndexes(trainedIndex faissIndexIV
 	var indexType faissIndexType
 	var validMerge bool
 
+	atomic.AddUint64(&v.stats.TotVecSectionMergesBegin, 1)
+	start := time.Now()
+
+	defer func() {
+		atomic.AddUint64(&v.stats.TotVecSectionMergeTime, uint64(time.Since(start)))
+		atomic.AddUint64(&v.stats.TotVecSectionMergesEnd, 1)
+	}()
```
maneuvertomars, line 494: This defer bumps MergesEnd (and merge time) even on error, while Persist/merge skip End on error, the two conventions contradict. Can we pick one (e.g. defer End always + a separate Errors counter) and use it everywhere?

maneuvertomars (commented):
We can have a look on these. Rest all looks good.

build.go
Thejas-bhat: I think given that this codepath still uses merge path to do a bunch of rewrite of the segment content in encryption usecases, we should still increment the associated relevant downstream stats. We're still tracking the flush operation itself as part of the TotPersist* stats, and realistically speaking when encryption isn't used, i don't think it bumps up any stats apart from writing the footer?

merge.go
Thejas-bhat: updated this in the latest PR

section_faiss_vector_index.go
Thejas-bhat: I don't think we should use Max naming convention for this since the stat still indicates the total number of vector fields in this index.

section_faiss_vector_index.go
Thejas-bhat: > panics in makeslice 

i think the panic only happens when the system has run out of memory? and the only case where the batch size can be large is on the fast merge path, for which we anyways size appropriately keeping the batch size in mind as well (this is something i've conveyed to tushar as well for sizing calculator).

in rest of the case, the batch is pretty small and i think doing a prealloc of a rough estimate for the entire batch (so alloc only once) is a benefit overall since we don't waste on realloc+GC? because we know that the total size needed for this batch is a multiple of the number of sub vectors

maybe we can remove this later on based on what you find, since i do see the benefit via the stats.

